PR #6, opened 2022-06-02
convert to grpc service

joostjager:
Convert multiplexer into a grpc service with an lnd-like api.


README.md
```diff
@@ -1,18 +1,14 @@
 # Lightning Multiplexer
 
-This repository contains the code for the `lnmux` library. The problem that `lnmux` solves is fail-over for incoming Lightning payments. BOLT11 invoices are bound to a single node. If that node goes down, the invoice is no longer payable. The same is true for the node or its peers running out of inbound liquidity.
+This repository contains the code for the `lnmux` service. The problem that `lnmux` solves is fail-over for incoming Lightning payments. BOLT11 invoices are bound to a single node. If that node goes down, the invoice is no longer payable. The same is true for the node or its peers running out of inbound liquidity.
```
qustavo, line 3: eventually, can this also be used as a go pkg?

cmd/multiplexer/server.go
```diff
@@ -0,0 +1,236 @@
...
+
+	if req.AmtMsat <= 0 {
+		return nil, errors.New("amt_msat not specified")
+	}
+
+	var descHash *lntypes.Hash
+	if len(req.DescriptionHash) > 0 {
+		hash, err := lntypes.MakeHash(req.DescriptionHash)
+		if err != nil {
+			return nil, fmt.Errorf("invalid desc hash: %v", err)
```
qustavo: `%w` to wrap the error

invoiceregistry_test.go
```diff
@@ -0,0 +1,148 @@
...
+
+	return ctx
+}
+
+func (r *registryTestContext) start() {
+	r.registry = NewRegistry(r.db, r.cfg)
+	require.NoError(r.t, r.registry.Start(context.Background()))
+}
+
+func (r *registryTestContext) stop() {
```
qustavo, line 66: rather than exposing a `stop()` and relying on the caller to defer it you could use `t.Cleanup()` on `newRegistryTestContext`

types/types.go
```diff
@@ -0,0 +1,56 @@
...
+	ChanID uint64
+
+	// HtlcID is the unique htlc index predominately assigned by links,
+	// though can also be assigned by switch in the case of locally-sourced
+	// payments.
+	HtlcID uint64
+}
+
+// String returns a string representation of the CircuitKey.
+func (k CircuitKey) String() string {
```
qustavo, line 54: why not pointer?

README.md
```diff
@@ -74,6 +74,16 @@ The steps below describe how to use this executable.
 
 * It should be possible to pay this invoice as long as at least one node in the cluster is available.
 
+* When the payment comes in, an event is generated. Events can be listened for via:
+  
+  `grpcurl -plaintext -v -d '{"hash":"EgR0u/TAjFXNGId9PEHVnryFl1sy/UwB2uN+YxC8S80="}' localhost:19090 lnmux.Multiplexer.SubscribeSingleInvoice`
```
qustavo: from a user perspective, I think having text hashes (hex encoded) are easier to use.

invoiceregistry.go
```diff
@@ -116,31 +131,174 @@ type InvoiceRegistry struct {
...
+		i.startInvoiceExpireTimer(hash, invoice.ExpiresAt)
+
+		i.logger.Debugw("Pending invoice",
+			"hash", hash, "expiresAt", invoice.ExpiresAt)
+	}
+
 	i.wg.Add(1)
-	go i.invoiceEventLoop()
+	go func() {
+		err := i.invoiceEventLoop(context.Background())
```
qustavo: shouldn't you use the ctx passed to `Start()`?

cmd/multiplexer/run.go
```diff
@@ -76,15 +104,61 @@ func runAction(c *cli.Context) error {
...
+	err = <-grpcErrChan
+	if err != nil {
+		return err
+	}
+
+	// Stop multiplexer.
+	log.Infof("Stopping multiplexer")
 	cancel()
-	return <-errChan
+	err = <-errChan
```
qustavo: is there's an error returned by returned `mux.Run()` we won't be able to catch it here until `sigint` channel returns a msg. Perhaps you should select over `sigint` and `errChan` chans

README.md
joostjager: yes, this is just a way to test the service. i think the proto should use native `bytes`, and then with grpcurl there is no other way to specify the key other than via base64 unforunately.

README.md
joostjager: I think it is still possible, but the intention of this pr is to switch the supported method to grpc only. At least initially. If it turns out that there is a significant group of users who need a golang library, we can reconsider.

cmd/multiplexer/server.go
joostjager: Done. Although I don't think it makes a difference for errors returned over grpc?

types/types.go
joostjager: For `String()` to work with calls like `Printf`, I think it is must be a value?

README.md
qustavo: you are right

types/types.go
qustavo: right, if you are not passing CircuitKey type as a pointer that makes sense.

cmd/multiplexer/run.go
joostjager: Fixed

invoiceregistry.go
joostjager: fixed

invoiceregistry_test.go
joostjager: Nice, done

PR #7, opened 2022-06-08
auto-settle

joostjager:
Adds the ability to add invoices in auto-settle mode.


invoiceregistry.go
```diff
@@ -409,6 +411,16 @@ func (i *InvoiceRegistry) invoiceEventLoop(ctx context.Context) error {
 				break
 			}
 
+			if state.autoSettle {
+				select {
```
qustavo: this select construct might be extracted now

invoiceregistry_test.go
```diff
@@ -170,3 +172,56 @@ func TestInvoiceExpiry(t *testing.T) {
...
+
+	update = <-updateChan
+	require.Equal(t, persistence.InvoiceStateSettleRequested, update.State)
+
+	update = <-updateChan
+	require.Equal(t, persistence.InvoiceStateSettled, update.State)
+
+	<-resolved
+
+	cancelUpdates()
```
qustavo, line 221: Invoked via `defer` for consistency (or `t.Cleanup`) ?

persistence/migrations/1_initial.up.sql
```diff
@@ -29,7 +29,9 @@ CREATE TABLE invoices
 
         "final_cltv_delta"      INTEGER         NOT NULL CHECK (final_cltv_delta > 0),
         "payment_addr"          BYTEA           NOT NULL UNIQUE CHECK (LENGTH(payment_addr) = 32),
-        "payment_request"       TEXT            NOT NULL CHECK (LENGTH(payment_request) > 0)
+        "payment_request"       TEXT            NOT NULL CHECK (LENGTH(payment_request) > 0),
```
qustavo, line 32: looks like `payment_request` should be part of a different pr

persistence/migrations/1_initial.up.sql
```diff
@@ -29,7 +29,9 @@ CREATE TABLE invoices
...
+
+        "auto_settle"           BOOLEAN         NOT NULL
```
qustavo, line 34: why aren't you creating a new migration?

qustavo: pre: missing down.sql migration

invoiceregistry.go
joostjager: done

invoiceregistry_test.go
joostjager: In `TestInvoiceExpiry`, I cancel them without a defer because there are multiple subscriptions that need to be cancelled before the test finishes. Therefore I also didn't use defer in this new test.

persistence/migrations/1_initial.up.sql
joostjager: Why? This is just adding a comma, it was already in master?

persistence/migrations/1_initial.up.sql
joostjager: I think the probability of this code being used in production is very low, so I figured we could forego the migrations for now until a bit more testing has been done.

persistence/migrations/1_initial.up.sql
qustavo: yes, my bad

invoiceregistry_test.go
qustavo: the fact that is not defered suggests that it will be cancelled somewhere in the middle of the test.
If you are going to call cancel at the end, why not prefer defer?

invoiceregistry.go
```diff
@@ -397,66 +403,62 @@ func (i *InvoiceRegistry) invoiceEventLoop(ctx context.Context) error {
...
+				return err
+			}
 
 		case req := <-i.cancelChan:
 			// Retrieve invoice.
 			state, ok := i.invoices[req.hash]
 			if !ok {
 				select {
-				case req.errChan <- errors.New("invoice not found"):
+				case req.errChan <- ErrInvoiceNotFound:
```
qustavo: can you use `sendResponse(errCh chan error, err error)` ?

invoiceregistry_test.go
joostjager: Seems better to me for consistency with the other test.

invoiceregistry.go
joostjager: nice follow up, done.

invoiceregistry_test.go
qustavo: ok

invoiceregistry.go
```diff
@@ -895,6 +915,43 @@ func (i *InvoiceRegistry) requestSettle(ctx context.Context,
...
+		)
+		i.notifyHodlSubscribers(key, htlcSettleResolution)
+	}
+
+	// TODO: Wait for final settle event from lnd. Unfortunately this
+	// event is not yet implemented.
+
+	// Mark invoice as settled.
+	if err := i.cdb.Settle(ctx, hash); err != nil {
+		return errors.New("cannot settle invoice in database")
```
qustavo: would it make sense to wrap the original error here to get more insight about what's failing?

invoiceregistry_test.go
```diff
@@ -156,3 +172,56 @@ func TestInvoiceExpiry(t *testing.T) {
...
+
+	preimage := c.preimage(1)
+	resolved := make(chan struct{})
+	c.registry.NotifyExitHopHtlc(&registryHtlc{
+		rHash:         preimage.Hash(),
+		amtPaid:       lnwire.MilliSatoshi(c.testAmt),
+		expiry:        100,
+		currentHeight: 0,
+		resolve: func(r HtlcResolution) {
+			c.logger.Infow("resolution", "res", r)
```
qustavo: do you need logging in test?

invoiceregistry_test.go
```diff
@@ -156,3 +172,56 @@ func TestInvoiceExpiry(t *testing.T) {
...
+			close(resolved)
+		},
+		payload: &testPayload{
+			amt:     lnwire.MilliSatoshi(c.testAmt),
+			payAddr: c.payAddr(1),
+		},
+	})
+
+	update = <-updateChan
```
qustavo, line 210: if there's an error and notifications are not sent, the test will wait 10 minutes (by default) until it times out (potentially blocking the CI). Would it be better to create a helper that reads updates from the channel with a much lower timeout?

Something like:

```go
assertFromChan := func(t *testing.T, ch chan<T>, state State) {
  t.Helper()

  select {
    case s := <- ch:
      require.Equal(t, state, s)
    case <- time.After(duration):
      t.Fatalf("timeout")
  }

File: invoiceregistry.go (hunk 6 above)
Comment (joostjager) [reply invoiceregistry.go]: yes, done

File: invoiceregistry_test.go (hunk 7 above)
Comment (joostjager) [reply invoiceregistry_test.go]: dev leftover, removed.

File: invoiceregistry_test.go (hunk 8 above)
Comment (joostjager) [reply invoiceregistry_test.go:210]: Added `test.Timeout` in a separate commit. This is a technique copied from `lnd`. That we you can still have a short per test timeout without cluttering the test code with selects everywhere.

Comment (qustavo) [reply invoiceregistry_test.go:210]: oh that's pretty clever

File: test/timeout.go [hunk 9]
```diff
@@ -0,0 +1,28 @@
...
+	"os"
+	"runtime/pprof"
+	"testing"
+	"time"
+)
+
+const testTimeout = 20 * time.Second
+
+// Timeout implements a test level timeout.
+func Timeout(t *testing.T) func() {
```
qustavo: you are passing `t` but is not being used.

joostjager: Fixed.

qustavo (approved):

PR #9, opened 2022-06-22
stateless invoices

joostjager:
Converts lnmux to always use stateless invoices. This means that nothing needs to be stored on disk when an invoice is generated.

As a result of this, the number of invoice (sub)states in the database is reduced as well. Only incoming payments for which settlement is requested are stored on disk.


mux_test.go
```diff
@@ -262,17 +256,9 @@ func TestMux(t *testing.T) {
...
-		InvoiceCreationData: invoice.InvoiceCreationData,
-		CreatedAt:           time.Now(),
-		PaymentRequest:      "payreq",
-		ID:                  2,
-		ExpiresAt:           time.Now().Add(time.Minute),
-	}))
-	expectUpdate(persistence.InvoiceStateOpen)
-
 	// Regenerate onion blob for new hash.
+	logger.Sugar().Infow("DEBUG POINT")
```
qustavo: ^

invoiceregistry.go
```diff
@@ -732,54 +659,73 @@ func (i *InvoiceRegistry) process(ctx context.Context, h *registryHtlc) error {
...
+		h.resolve(NewFailResolution(ResultInvoiceNotFound))
 
-			h.resolve(NewFailResolution(ResultInvoiceNotFound))
+		return nil
+	}
 
-			return nil
-		}
+	// Extract payment parameters.
+	expiry := time.Unix(int64(byteOrder.Uint64(paymentAddr[16:24])), 0)
```
qustavo: I'd rather create a `PaymentAddr` type that receives amt and expiry as construction params and returns them via `Expiry` and `Amount` methods.
This way all the encoding/decoding logics remain on a single place

cmd/lnmuxd/config.go
```diff
@@ -19,6 +19,9 @@ type Config struct {
 
 	// IdentityKey is the private key that is used to sign invoices.
 	IdentityKey string `yaml:"identityKey"`
+
+	// AutoSettle indicates that payments should be accepted automatically.
+	AutoSettle bool `yaml:"autoSettle"`
```
qustavo, line 24: I wonder if you could keep 1 bit in the payment addr to indicate if the invoice should auto-settle

invoiceregistry.go
```diff
@@ -721,7 +608,47 @@ type registryHtlc struct {
...
+
+	return nil
+}
+
 func (i *InvoiceRegistry) process(ctx context.Context, h *registryHtlc) error {
+	logger := i.logger.With("hash", h.rHash)
+
+	// First try to resolve via the database, in case this is a replay.
+	err := i.resolveViaDb(ctx, h)
+	if err != types.ErrInvoiceNotFound {
```
qustavo: I think this error handling should be absorbed by `.resolveViaDb`

invoiceregistry.go
```diff
@@ -732,54 +659,73 @@ func (i *InvoiceRegistry) process(ctx context.Context, h *registryHtlc) error {
...
+			acceptedHtlcs: make(map[types.CircuitKey]*InvoiceHTLC),
+			expiry:        expiry,
 		}
 
-		// Otherwise fail the htlc.
-		h.resolve(NewFailResolution(ResultInvoiceNotOpen))
+		i.invoices[h.rHash] = state
+	} else {
+		// Sanity check that the total amount and expiry time are identical.
+		if amtMsat != uint64(state.invoice.Value) ||
```
qustavo: Is this even possible?

mux_test.go
joostjager: fixed

invoiceregistry.go
joostjager: Done

cmd/lnmuxd/config.go
joostjager: It is possible, but I don't know if there is a valid use case for this. Seems more like a system-level choice to me whether to auto-settle.

invoiceregistry.go
joostjager: Done, used additional `resolved` return value.

invoiceregistry.go
joostjager: Normally not. Unless a bug has occurred or our private key was compromised and used to fabricate invalid payment addresses.

cmd/lnmuxd/config.go
qustavo: agreed

invoiceregistry.go
qustavo: If any of those happen, shouldn't we `.Fatal`? IMO it's fair to stop the execution in any of the cases you describe (bug or pk compromised)

invoiceregistry_test.go
```diff
@@ -128,70 +125,42 @@ func TestInvoiceExpiry(t *testing.T) {
...
-	update = <-updateChan2
-	require.Equal(t, persistence.InvoiceStateOpen, update.State)
-	cancel2()
-
-	// Stop the registry.
-	c.stop()
+	invoice, preimage := c.createInvoice(1, time.Second)
 
 	// Wait for the invoice to expire.
 	time.Sleep(2 * time.Second)
```
qustavo: pre: you could use smaller durations in order to make tests run faster?

invoiceregistry_test.go
```diff
@@ -128,70 +125,42 @@ func TestInvoiceExpiry(t *testing.T) {
...
+		payload: &testPayload{
+			amt:     lnwire.MilliSatoshi(c.testAmt),
+			payAddr: invoice.PaymentAddr,
+		},
+	})
 
-	case <-time.After(200 * time.Millisecond):
-	}
-	cancel3()
+	require.IsType(t, &HtlcFailResolution{}, <-resolved)
```
qustavo: you could also add: 

```go
	require.Equal(t, ResultInvoiceExpired, resolution.(*HtlcFailResolution).Outcome)
```

invoiceregistry.go
joostjager: My internal guideline is to only panic when something really can't happen. A compromised key is valid enough to not panic (literally) I think.

Wondering what the correct way to handle this is though.

invoiceregistry_test.go
joostjager: I don't really want to make them super small, to prevent flakes when the CI runner proves to be slow.

invoiceregistry.go
qustavo: As the operator of the software, if a key gets compromised I'd like get notified and probably stop creating invoices with that key.

invoiceregistry_test.go
qustavo: but as long as time.Sleep's duration is higher than createInvoices' we should be ok isnt'?

cmd/lnmuxd/config.go
aakselrod-nydig, line 24: I like this change to set consistent policy across invoices.

cmd/lnmuxd/lnmux_proto/lnmux.proto
```diff
@@ -19,11 +19,6 @@ message AddInvoiceRequest {
     string description = 2;
     bytes description_hash = 3;
     int64 expiry_secs = 4;
-
-    // If set to true, invoices will be settled automatically when a payment to
-    // it comes in. It is not necessary or possible to request settlement
-    // manually.
-    bool auto_settle = 5;
```
aakselrod-nydig, line 26: Less data in each request, and less code to spec it! 🎉

cmd/lnmuxd/lnmux_proto/lnmux.proto
```diff
@@ -38,23 +33,12 @@ message SubscribeSingleInvoiceRequest {
...
-    }
-
-    enum CancelledReason {
-      REASON_NONE = 0;
-      REASON_EXPIRED = 1;
-      REASON_ACCEPT_TIMEOUT = 2;
-      REASON_EXTERNAL = 3;
+      STATE_ACCEPTED = 0;
+      STATE_SETTLE_REQUESTED = 1;
+      STATE_SETTLED = 2;
```
aakselrod-nydig, line 38: Much simpler, no need to store any info about invoices that don't attempt settlement!

cmd/lnmuxd/server.go
```diff
@@ -136,8 +111,7 @@ func (s *server) SubscribeSingleInvoice(req *lnmux_proto.SubscribeSingleInvoiceR
 			}
 
 			err := subscription.Send(&lnmux_proto.SubscribeSingleInvoiceResponse{
-				State:           marshallInvoiceState(update.State),
-				CancelledReason: marshallCancelledReason(update.CancelledReason),
+				State: marshallInvoiceState(update.State),
```
aakselrod-nydig, line 114: Simpler handling! 🎉

cmd/lnmuxd/server.go
```diff
@@ -170,7 +144,6 @@ func (s *server) AddInvoice(ctx context.Context,
 
 	// Create the invoice.
 	expiry := time.Duration(req.ExpirySecs) * time.Second
-	expiryTime := time.Now().Add(expiry)
```
aakselrod-nydig, line 173: Good refactor to move this to the `InvoiceCreator`.

cmd/lnmuxd/server.go
```diff
@@ -180,21 +153,6 @@ func (s *server) AddInvoice(ctx context.Context,
...
-		PaymentRequest:      invoice.PaymentRequest,
-		ExpiresAt:           expiryTime,
-		AutoSettle:          req.AutoSettle,
-	}
-
-	err = s.registry.NewInvoice(creationData)
-	if err != nil {
-		return nil, err
-	}
-
```
aakselrod-nydig, line 197: :tada: :fire: no more storing invoices on creation!

invoiceregistry.go
```diff
@@ -83,6 +89,10 @@ func (i *invoiceState) totalSetAmt() int {
 	return total
 }
 
+func (i *invoiceState) isSetComplete() bool {
+	return i.totalSetAmt() == int(i.invoice.Value)
```
aakselrod-nydig, line 93: Should there be any handling for when the total set amount is greater than the invoice value (sender sends too much money)? If so, maybe a TODO here?

invoiceregistry.go
```diff
@@ -161,45 +169,7 @@ func NewRegistry(cdb *persistence.PostgresPersister,
...
-		hash := invoice.PaymentPreimage.Hash()
-
-		i.invoices[hash] = state
-		i.startInvoiceExpireTimer(hash, invoice.ExpiresAt)
-
-		i.logger.Debugw("Pending invoice",
-			"hash", hash, "expiresAt", invoice.ExpiresAt)
-	}
-
-	err = i.invoiceEventLoop(ctx)
```
aakselrod-nydig, line 202: :fire:

invoiceregistry.go
```diff
@@ -345,17 +305,9 @@ func (i *InvoiceRegistry) invoiceEventLoop(ctx context.Context) error {
...
 				}
 
-			case *invoiceExpiredEvent:
-				err := i.failInvoice(
-					ctx, event.hash, persistence.CancelledReasonExpired,
-				)
-				if err != nil {
-					return err
-				}
-
```
aakselrod-nydig, line 355: :fire: :zap:

invoiceregistry.go
```diff
@@ -368,32 +320,6 @@ func (i *InvoiceRegistry) invoiceEventLoop(ctx context.Context) error {
...
-			i.invoices[hash] = state
-
-			// Notify subscriber of new invoice.
-			i.subscriptionManager.notifySubscribers(hash, InvoiceUpdate{
-				State: persistence.InvoiceStateOpen,
-			})
-
-			i.startInvoiceExpireTimer(hash, invoice.ExpiresAt)
-
```
aakselrod-nydig, line 396: :zap: :fire: :zap:

invoiceregistry.go
```diff
@@ -460,11 +386,6 @@ func (i *InvoiceRegistry) invoiceEventLoop(ctx context.Context) error {
 				break
 			}
 
-			// Mark invoice as failed.
-			if err := i.cdb.Fail(ctx, req.hash, persistence.CancelledReasonExternal); err != nil {
-				return errors.New("cannot fail invoice in database")
-			}
-
```
aakselrod-nydig, line 467: :zap:

invoiceregistry.go
```diff
@@ -478,15 +399,6 @@ func (i *InvoiceRegistry) invoiceEventLoop(ctx context.Context) error {
...
 
-			// Notify subscriber of settled invoice.
-			i.subscriptionManager.notifySubscribers(
-				req.hash,
-				InvoiceUpdate{
-					State:           persistence.InvoiceStateCancelled,
-					CancelledReason: persistence.CancelledReasonExternal,
-				},
-			)
-
```
aakselrod-nydig, line 489: :zap: :fire:

invoiceregistry.go
```diff
@@ -528,10 +440,12 @@ func (i *InvoiceRegistry) addSubscriber(ctx context.Context,
 
 	invoiceState, ok := i.invoices[hash]
 	if ok {
-		update.State = persistence.InvoiceStateOpen
-		if len(invoiceState.acceptedHtlcs) > 0 {
-			update.State = persistence.InvoiceStateAccepted
+		// No event for partially accepted invoices.
+		if !invoiceState.isSetComplete() {
```
aakselrod-nydig, line 444: In this case, do we want an event when the total amount sent is greater than the requested amount in the invoice? Maybe a TODO here for that as well?

invoiceregistry.go
```diff
@@ -568,53 +484,19 @@ func (i *InvoiceRegistry) failInvoice(ctx context.Context,
 		return nil
 	}
 
-	// Don't expire invoices that are already accepted.
-	setComplete := state.totalSetAmt() == int(state.invoice.Value)
-	if reason == persistence.CancelledReasonExpired && setComplete {
-		return nil
-	}
-
```
aakselrod-nydig, line 576: :zap:

invoiceregistry.go
```diff
@@ -568,53 +484,19 @@ func (i *InvoiceRegistry) failInvoice(ctx context.Context,
...
 	for key := range state.acceptedHtlcs {
 		i.notifyHodlSubscribers(key, NewFailResolution(ResultInvoiceExpired))
 	}
 
-	// Mark invoice as expired in the database.
-	err := i.cdb.Fail(ctx, hash, reason)
-	if err != nil {
-		return err
-	}
-
```
aakselrod-nydig, line 587: :zap:

invoiceregistry.go
```diff
@@ -568,53 +484,19 @@ func (i *InvoiceRegistry) failInvoice(ctx context.Context,
...
 	// Remove from memory because invoice is no longer open.
 	delete(i.invoices, hash)
 
-	// Notify subscriber.
-	i.subscriptionManager.notifySubscribers(hash, InvoiceUpdate{
-		State:           persistence.InvoiceStateCancelled,
-		CancelledReason: reason,
-	})
-
```
aakselrod-nydig, line 596: :zap:

invoiceregistry.go
```diff
@@ -568,53 +484,19 @@ func (i *InvoiceRegistry) failInvoice(ctx context.Context,
...
-			releaseTime: releaseTime,
-		},
-	}
-
-	i.logger.Debugw("Scheduling auto-release for invoice",
-		"hash", hash, "releaseTime", releaseTime)
-
-	i.autoReleaseHeap.Push(event)
-}
-
```
aakselrod-nydig, line 617: :zap:

stateless_data.go
```diff
@@ -0,0 +1,83 @@
...
+		return nil, err
+	}
+
+	return &statelessData{
+		amtMsat:     amtMsat,
+		expiry:      expiry,
+		preimage:    paymentPreimage,
+		paymentAddr: addr,
+	}, nil
+}
```
aakselrod-nydig, line 59: ⚡ 🔥

stateless_data.go
```diff
@@ -0,0 +1,83 @@
...
+	expiry := time.Unix(int64(byteOrder.Uint64(paymentAddr[16:24])), 0)
+	amtMsat := byteOrder.Uint64(paymentAddr[24:32])
+
+	return &statelessData{
+		amtMsat:     int64(amtMsat),
+		expiry:      expiry,
+		paymentAddr: paymentAddr,
+		preimage:    paymentPreimage,
+	}, nil
+}
```
aakselrod-nydig, line 83: :zap: :fire:

aakselrod-nydig (approved):
This looks great and simplifies the code a lot! I did leave a couple of comments about potentially handling a set of HTLCs that overpays an invoice, which could be especially useful in the case of donations or overpayments for the sake of reliable routing. If you want to make a TODO, I would like to implement a test for this condition and the handling of it as a first issue for me?

invoiceregistry.go
joostjager: Yes I get that. But wondering what the best communication path is for that. I could introduce prometheus metrics in this repo for example, if that is considered a good option.

What would you do in the context of this PR?

invoiceregistry.go
joostjager: The current implementation rejects overpayments intentionally, to avoid having to deal with the unexpected extra sats. An application connected to multiplexer needs to be able to deal with overpayment too, it needs to be accounted for properly, etc.

It could be a config setting though, whether to accept overpayments or not.

Or when auto-settle is disabled, the application could inspect the actually paid value during processing of the `ACCEPTED` event and reject if it doesn't want to handle overpayments.

Another variation is to encode 'accept-overpayment' as a flag in the payment address too, so it can be different per invoice.

Then there is also the case of zero valued invoices, which are not yet supported either.

A different, even more minimalistic approach could be to not encode expiry and amount in the payment address at all. Just pass every payment on to the application for deciding whether to settle the payment or not. Perhaps that's a bit too extreme though. And also it won't work with auto-settle.

stateless_data.go
joostjager: These are a lot of lightning fires

invoiceregistry_test.go
joostjager: Added

invoiceregistry_test.go
joostjager: Ah yes, that's true in this case. Fixed.

invoiceregistry.go
joostjager: As mentioned offline, I am suggesting to add monitoring in a follow up pr. Think a bit deeper about it and also add other metrics that are relevant.

Jawshua: Prometheus metrics are pretty much a requirement if we want to actually run this.

qustavo (approved):

invoiceregistry.go
joostjager: https://github.com/bottlepay/lnmux/pull/42

PR #12, opened 2022-07-12
move proto code to its own module

joostjager:
This allows users to import just the proto code and not all of the lnmux dependencies. In particular lnd as a dependency can be problematic and conflicting with a specific version that the application depends on.


aakselrod-nydig:
⚡  LGTM, great move!

PR #13, opened 2022-07-12
return structured grpc errors

joostjager:


aakselrod-nydig (approved):
Looks good! Is the plan to do the other two in invoiceregistry.go in this PR as well?

joostjager:
Which other two ones do you mean?

Merging this already now.

Issue #10, opened 2022-07-06
Add test for overpaid invoices


aakselrod-nydig:
There are times when a sender may overpay an invoice. As we consider what to do with such overpayments, it would be helpful to include a test for their handling (expected to fail until handling is added). In addition, it would also be helpful to check the code coverage of our tests as we move toward putting lnmux into production with real money on mainnet.

aakselrod-nydig:
Fixed by #18

PR #18, opened 2022-07-14
Overpayment test

aakselrod:
This PR is a response to #10. In it:

- a new test is added which checks that overpaying HTLCs are rejected, including in combination, but an invoice is still accepted when the MPP set is complete
- a code coverage percentage is generated for each package and shown in the unit test logs

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,6 +26,6 @@ jobs:
       - name: Run tests
         # Runs a script that creates a PostgreSQL client,
         # populates the client with data, and retrieves data
-        run: go test ./...
+        run: go test -cover ./...
         env:
           LNMUX_TEST_DB_DSN: "postgres://bottle_test:bottle_test@postgres/bottle_test?sslmode=disable"
--- a/invoiceregistry_test.go
+++ b/invoiceregistry_test.go
@@ -131,6 +131,23 @@ func (r *registryTestContext) subscribeAccept() (chan lntypes.Hash, func()) {
 	return updateChan, cancel
 }
 
+func (r *registryTestContext) checkHtlcFailed(resolution HtlcResolution,
+	result FailResolutionResult) {
+
+	require.IsType(r.t, &HtlcFailResolution{}, resolution)
+	require.Equal(r.t, result, resolution.(*HtlcFailResolution).Outcome)
+}
+
+func (r *registryTestContext) checkHtlcSettled(resolution HtlcResolution) {
+	require.IsType(r.t, &HtlcSettleResolution{}, resolution)
+}
+
+func (r *registryTestContext) checkUpdate(update InvoiceUpdate,
+	expected persistence.InvoiceState) {
+
+	require.Equal(r.t, expected, update.State)
+}
+
 func TestInvoiceExpiry(t *testing.T) {
 	defer test.Timeout()()
 
@@ -158,9 +175,7 @@ func TestInvoiceExpiry(t *testing.T) {
 		},
 	})
 
-	resolution := <-resolved
-	require.IsType(t, &HtlcFailResolution{}, resolution)
-	require.Equal(t, ResultInvoiceExpired, resolution.(*HtlcFailResolution).Outcome)
+	c.checkHtlcFailed(<-resolved, ResultInvoiceExpired)
 }
 
 func TestAutoSettle(t *testing.T) {
@@ -193,14 +208,11 @@ func TestAutoSettle(t *testing.T) {
 	acceptedInvoice := <-acceptChan
 	require.Equal(t, preimage.Hash(), acceptedInvoice)
 
-	update := <-updateChan
-	require.Equal(t, persistence.InvoiceStateAccepted, update.State)
+	c.checkUpdate(<-updateChan, persistence.InvoiceStateAccepted)
 
-	update = <-updateChan
-	require.Equal(t, persistence.InvoiceStateSettleRequested, update.State)
+	c.checkUpdate(<-updateChan, persistence.InvoiceStateSettleRequested)
 
-	update = <-updateChan
-	require.Equal(t, persistence.InvoiceStateSettled, update.State)
+	c.checkUpdate(<-updateChan, persistence.InvoiceStateSettled)
 
 	<-resolved
 
@@ -239,16 +251,13 @@ func TestSettle(t *testing.T) {
 		},
 	})
 
-	update := <-updateChan
-	require.Equal(t, persistence.InvoiceStateAccepted, update.State)
+	c.checkUpdate(<-updateChan, persistence.InvoiceStateAccepted)
 
 	require.NoError(t, c.registry.RequestSettle(hash))
 
-	update = <-updateChan
-	require.Equal(t, persistence.InvoiceStateSettleRequested, update.State)
+	c.checkUpdate(<-updateChan, persistence.InvoiceStateSettleRequested)
 
-	update = <-updateChan
-	require.Equal(t, persistence.InvoiceStateSettled, update.State)
+	c.checkUpdate(<-updateChan, persistence.InvoiceStateSettled)
 
 	<-resolved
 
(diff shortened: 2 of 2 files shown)


invoiceregistry.go
```diff
@@ -1009,7 +1013,7 @@ func (i *InvoiceRegistry) markSettleRequested(ctx context.Context,
 			"setTotal", setTotal,
 			"invoiceValue", state.invoice.Value)
 
-		return errors.New("set no longer complete")
+		return ErrSetNoLongerComplete
```
joostjager: I sort of forgot about this case where some of the htlcs have been released, but not all. Would be good to also convert this error to a grpc error in `server.SettleInvoice`.

.github/workflows/test.yml
```diff
@@ -26,6 +26,6 @@ jobs:
       - name: Run tests
         # Runs a script that creates a PostgreSQL client,
         # populates the client with data, and retrieves data
-        run: go test ./...
+        run: go test -coverprofile=coverage.out ./...
```
joostjager: How can this output be viewed?

mux_test.go
```diff
@@ -288,6 +288,34 @@ func TestMux(t *testing.T) {
...
+
+	// Regenerate onion blob for new hash.
+	onionBlob = genOnion()
+
+	htlcChan1 <- receiveHtlc(30, 20001)
+
+	// Expect failure for MPP overpayment
+	// TODO(aakselrod): decrypt and check failure message
+	expectResponse(<-responseChan1, 30, routerrpc.ResolveHoldForwardAction_FAIL)
+	require.ErrorIs(t, registry.RequestSettle(testHash), ErrSetNoLongerComplete)
```
joostjager: Maybe something is not right. I'd expect this error only in the case that one or more htlcs of a set have been release (htlc release timeout), but there are still some left.

Overpayment should hit https://github.com/bottlepay/lnmux/blob/fd33bf0d8627d325586f7f844ed37e7cfe7b1615/invoiceregistry.go#L866 and I'd expect the set to be removed from memory.

joostjager: Yes, looked more closely at it and we definitely have a bug. The idea is that we only keep invoices in memory (`InvoiceRegistry.invoices`) that have a non-empty htlc set. But the check above and a few others in that same area leave an empty set.

joostjager: Did a bigger refactor to be more certain of the correctness: https://github.com/bottlepay/lnmux/pull/19

aakselrod-nydig: With the refactor, I was able to make a better test case.

.github/workflows/test.yml
aakselrod-nydig: Right now, the output file is temporary (deleted on container destruction) and the only visible output is in the logs, showing a percentage per package that has tests in it (mainly just the `lnmux` and `persistence` packages. I would like to get this reporting to something with easy-to-read output, perhaps CodeCov or shields.io?

invoiceregistry.go
aakselrod-nydig: This should be addressed in https://github.com/bottlepay/lnmux/pull/18/commits/31d4088731b5a33ed0332e6f8f0b05b493848fbb now

mux_test.go
```diff
@@ -314,7 +315,7 @@ func TestMux(t *testing.T) {
 	// Expect failure for MPP overpayment
 	// TODO(aakselrod): decrypt and check failure message
 	expectResponse(<-responseChan1, 30, routerrpc.ResolveHoldForwardAction_FAIL)
-	require.ErrorIs(t, registry.RequestSettle(testHash), ErrSetNoLongerComplete)
```
joostjager: The commits that rewrite the test can be squashed.

.github/workflows/test.yml
joostjager: 71%, not bad. It would be nice to view the coverage somewhere, but for me personally it isn't the highest priority because vscode also shows coverage when running tests locally.

cmd/lnmuxd/server.go
```diff
@@ -246,6 +246,11 @@ func (s *server) SettleInvoice(ctx context.Context,
 			codes.Unavailable, lnmux.ErrAutoSettling.Error(),
 		)
 
+	case err == lnmux.ErrSetNoLongerComplete:
+		return nil, status.Error(
+			codes.FailedPrecondition, lnmux.ErrSetNoLongerComplete.Error(),
+		)
```
joostjager: :+1:

mux_test.go
```diff
@@ -288,6 +289,52 @@ func TestMux(t *testing.T) {
...
+
+		updateChan <- update
+	})
+	require.NoError(t, err)
+
+	// Regenerate onion blob for new hash.
+	onionBlob = genOnion()
+
+	// Expect failure for MPP overpayment
+	// TODO(aakselrod): decrypt and check failure message
```
joostjager: You want to do this in this PR?

An alternative would be to test overpayment in the invoice registry test, because it doesn't require decoding.

mux_test.go
```diff
@@ -288,6 +289,52 @@ func TestMux(t *testing.T) {
...
+	require.ErrorIs(t, registry.RequestSettle(testHash), types.ErrInvoiceNotFound)
+	expectResponse(<-responseChan1, 31, routerrpc.ResolveHoldForwardAction_FAIL)
+
+	// Now a pair of payments that should succeed, replaying the first and
+	// sizing the second correctly
+	htlcChan1 <- receiveHtlc(31, 15000)
+	htlcChan2 <- receiveHtlc(33, 5000)
+	expectAccept(testHash)
+	expectUpdate(persistence.InvoiceStateAccepted)
+	require.NoError(t, registry.RequestSettle(testHash))
```
joostjager: A bit of breathing space wouldn't be bad here (newline above?)

mux_test.go
```diff
@@ -288,6 +289,52 @@ func TestMux(t *testing.T) {
...
+
+	// Now do the same with a two-part payment
+	htlcChan1 <- receiveHtlc(31, 15000)
+	htlcChan2 <- receiveHtlc(32, 6000)
+	expectResponse(<-responseChan2, 32, routerrpc.ResolveHoldForwardAction_FAIL)
+	require.ErrorIs(t, registry.RequestSettle(testHash), types.ErrInvoiceNotFound)
+	expectResponse(<-responseChan1, 31, routerrpc.ResolveHoldForwardAction_FAIL)
+
+	// Now a pair of payments that should succeed, replaying the first and
+	// sizing the second correctly
```
joostjager: What is the exact scenario with this replay? Maybe lnd crashed and didn't receive the previous failure for number 31?

aakselrod-nydig: I've eliminated this and there's no replay in the new test.

mux_test.go
aakselrod-nydig: I've eliminated this test.

mux_test.go
aakselrod-nydig: I tried the latter and it came out much cleaner!

.github/workflows/test.yml
aakselrod-nydig: I've changed this to only output to logs, not to a file. No difference in the end, but then there's no question about what happens with the output: it's only visible in the logs, in case it's wanted.

mux_test.go
aakselrod-nydig: I've eliminated this test altogether.

invoiceregistry.go
```diff
@@ -939,7 +943,7 @@ func (i *InvoiceRegistry) markSettleRequested(ctx context.Context,
 			"setTotal", setTotal,
 			"invoiceValue", set.value())
 
-		return errors.New("set no longer complete")
+		return ErrSetNoLongerComplete
```
joostjager: I am in doubt now if it isn't an abstraction break to expose this at all. For the perspective of the application using lnmux, everything happens on the level of payments. But now here, we are exposing a detail about the htlc set.

Also I realized that if you try to settle _before_ the set is complete, you'll get this error too (I think?). It should then be `ErrSetNotYetComplete`. So maybe best to avoid this completely?

invoiceregistry_test.go
```diff
@@ -299,6 +299,106 @@ func TestCancel(t *testing.T) {
...
+	require.IsType(t, &HtlcFailResolution{}, resolution)
+	require.Equal(t, ResultHtlcSetOverpayment, resolution.(*HtlcFailResolution).Outcome)
+
+	// We should have neither acceptance nor update here.
+	select {
+	case <-acceptChan:
+		require.FailNow(t, "Invoice accepted incorrectly")
+	case <-updateChan:
+		require.FailNow(t, "Update generated incorrectly")
+	default:
```
joostjager: Maybe a small `<-time.After(...)` here, in case the implementation is such that the resolution is sent first? Otherwise it could be racy.

invoiceregistry_test.go
```diff
@@ -299,6 +299,106 @@ func TestCancel(t *testing.T) {
...
+	require.IsType(t, &HtlcFailResolution{}, resolution)
+	require.Equal(t, ResultHtlcSetOverpayment, resolution.(*HtlcFailResolution).Outcome)
+
+	// We should not have a second resolution, an update, or an accept here.
+	select {
+	case <-acceptChan:
+		require.FailNow(t, "Invoice accepted incorrectly")
+	case <-updateChan:
+		require.FailNow(t, "Update generated incorrectly")
+	case <-resolved:
```
joostjager, line 386: Why another read here?

invoiceregistry_test.go
```diff
@@ -299,6 +299,106 @@ func TestCancel(t *testing.T) {
...
+
+	update = <-updateChan
+	require.Equal(t, persistence.InvoiceStateSettled, update.State)
+
+	// Both of the HTLCs should now resolve as settled.
+	resolution = <-resolved
+	require.IsType(t, &HtlcSettleResolution{}, resolution)
+
+	resolution = <-resolved
+	require.IsType(t, &HtlcSettleResolution{}, resolution)
```
joostjager: Nice test. Maybe it is possible to move a few common code blocks to the test context for reuse?

invoiceregistry_test.go
aakselrod-nydig: It's expected to be empty to ensure only one HTLC (ID 3) is resolved as checked above, and that there are no other resolved HTLCs. I could add a `<-time.After(...)` here as well instead of default, to make sure.

invoiceregistry_test.go
aakselrod-nydig: Sounds good, I did it that way at first but then took it out. I'll add a 1s delay here.

invoiceregistry.go
aakselrod-nydig: I'll take it out for now. It's unnecessary now that we're testing overpayment in the invoice registry tests.

invoiceregistry_test.go
aakselrod-nydig: I've done this and also added another check of the `resolved` channel to ensure that there's no other resolution. I also had to change the delay to 100ms to ensure that the HTLC doesn't time out.

invoiceregistry_test.go
aakselrod-nydig: As above, changed to 100ms timeout.

invoiceregistry_test.go
aakselrod-nydig: I've extracted three utility functions into the test context, and used them in the other tests as well.

invoiceregistry_test.go
```diff
@@ -131,6 +131,21 @@ func (r *registryTestContext) subscribeAccept() (chan lntypes.Hash, func()) {
 	return updateChan, cancel
 }
 
+func (r *registryTestContext) checkHtlcFailed(resolution HtlcResolution,
+	result FailResolutionResult) {
+	require.IsType(r.t, &HtlcFailResolution{}, resolution)
```
joostjager, line 137: We usually do a new line here, to separate the function signature from the body

invoiceregistry_test.go
```diff
@@ -299,6 +306,101 @@ func TestCancel(t *testing.T) {
...
+	// Test with full overpayment in a single HTLC.
+	notifyHtlc(1, c.testAmt+1)
+
+	c.checkHtlcFailed(<-resolved, ResultHtlcSetOverpayment)
+
+	// We should have neither acceptance nor update here.
+	select {
+	case <-acceptChan:
+		require.FailNow(t, "Invoice accepted incorrectly")
+	case <-updateChan:
```
joostjager, line 356: We also insert newlines to separate the case statements. There is no style guide unfortunately or a linter that enforces this. In general we adhere to the lnd coding style.

invoiceregistry_test.go
joostjager: Ok, to check that there is exactly one resolution, got it.

invoiceregistry_test.go
joostjager: Looks good

joostjager (approved):

PR #19, opened 2022-07-14
refactor htlcset

joostjager:
Various bug fixes and refactors, see commit messages.


aakselrod-nydig (approved):
This looks to me to match the updated state diagram in #21 except as we discussed there re. state change from Accepted to Expired (done on HTLC expiration rather than invoice expiration unless settlement is requested).

PR #21, opened 2022-07-14
update state diagram

joostjager:


aakselrod-nydig (approved):
LGTM

joostjager:
Actually, I think it is not correct. I don't believe it goes from accepted to expired.

aakselrod-nydig:
It looks like it DOES go from accepted to expired, when the accept timer expires rather than when the invoice expires?

https://github.com/bottlepay/lnmux/blob/e8dc6c8c047944b6a1ec30207149c3826d75bba4/invoiceregistry.go#L327

joostjager:
Yes, it is not the invoice expiry. There is already an arrow for accept timeout in the diagram.

PR #26, opened 2022-07-19
rename to SubscribeInvoiceAccepted

joostjager:
Even though it isn't really an invoice that is accepted, it is more in line
with the general terminology.


aakselrod-nydig (approved):
Makes sense, especially with the new optimized invoice lifecycle. LGTM!

PR #27, opened 2022-07-19
build: restrict push trigger

joostjager:
To prevent duplicate workflows being kicked off for pull requests.


aakselrod-nydig (approved):
Nice catch!

PR #30, opened 2022-07-20
add set id

joostjager:
introduces a set id that is communicated as part of the invoice acceptance event. This set id needs to be passed in to cancel or settle an invoice.

The purpose is to make it explicit which set the caller acts on. If the caller is slow, a set may be released again and a new set accepted. Especially if we accept over-payment, the caller may not be aware that they are accepting a different payment amount.

Builds on #29


set_id.go
```diff
@@ -0,0 +1,42 @@
...
+
+		return sortedKeys[a].ChanID < sortedKeys[b].ChanID
+	})
+
+	keysBytes := make([]byte, len(sortedKeys)*16)
+	idx := 0
+	for _, key := range sortedKeys {
+		byteOrder.PutUint64(keysBytes[idx:], key.ChanID)
+		byteOrder.PutUint64(keysBytes[idx+8:], key.HtlcID)
+		idx += 8
```
aakselrod-nydig, line 36: I think this should be a 16. I added a `fmt.Println(keysBytes)` after this loop, and it looks like the following output when I run the test:

```
=== RUN   TestSetID
[0 0 0 0 0 0 0 1 0 0 0 0 0 0 0 2 0 0 0 0 0 0 0 4 0 0 0 0 0 0 0 0]
[0 0 0 0 0 0 0 1 0 0 0 0 0 0 0 2 0 0 0 0 0 0 0 4 0 0 0 0 0 0 0 0]
[0 0 0 0 0 0 0 1 0 0 0 0 0 0 0 2 0 0 0 0 0 0 0 5 0 0 0 0 0 0 0 0]
--- PASS: TestSetID (0.00s)
```

When this index is incremented by 16 in each loop, we get the following output:

```
[0 0 0 0 0 0 0 1 0 0 0 0 0 0 0 3 0 0 0 0 0 0 0 2 0 0 0 0 0 0 0 4]
[0 0 0 0 0 0 0 1 0 0 0 0 0 0 0 3 0 0 0 0 0 0 0 2 0 0 0 0 0 0 0 4]
[0 0 0 0 0 0 0 1 0 0 0 0 0 0 0 3 0 0 0 0 0 0 0 2 0 0 0 0 0 0 0 5]
```

joostjager: Nice catch

PR #35, opened 2022-07-22
refactor to use errgroup

joostjager:
or more structured error handling and goroutine waiting. Also prepares for additional goroutines to be added such as a metrics endpoint.


aakselrod-nydig (approved):
This looks great! Fewer LoC and more readable. I didn't know about `errgroup` but I will in the future!

PR #41, opened 2022-07-25
Add settled state

### LINKED ISSUE #33: Revive invoice settled state
In https://github.com/bottlepay/lnmux/pull/29, we remove the invoice settled state because it is fake and not actually backed by any check on lnd.

A reliable implementation of the settled state can only be realized after https://github.com/lightningnetwork/lnd/pull/6517 is done. This is scheduled for lnd 0.16, so will take another 3-6 months.

An intermediary solution could be to define the settled state as follows:

"For every htlc of the set, a `settle` htlc interceptor action has been sent to the corresponding lnd instance and an OK result was returned".

This makes lnmux safer than it currently is, because a connecting application will at least wait for the settle messages to be delivered. There is still an uptime requirement for lnmux though, because lnd isn't persisting the settle message.

joostjager:
Makes multiplexer safer to use by providing an endpoint to wait for the settle message to be delivered to all involved lnd instances.

fixes #33


Torakushi:
Sounds good to me! Looking forward to LND PR for even more security!

aakselrod-nydig (approved):
LGTM so far!

Issue #23, opened 2022-07-18
Assess quality of logging and improve if necessary


joostjager:
Making sure that in case of issues we've got all relevant information captured.

Issue #15, opened 2022-07-12
Add prometheus metrics


joostjager:
Both for monitoring regular operation (number of htlcs settled, failed, etc) as well as certain error cases that are worth keeping an eye on. One of those is a mismatch of the rederived preimage and the payment hash.

Torakushi:
In case of problem, it could be good to add some prometheus metrics for our grpc_server:

https://github.com/grpc-ecosystem/go-grpc-prometheus

joostjager:
I think we already use that in lnmux?

Torakushi:
I missed this sorry...

Torakushi:
Coule be nice to have a gauge indicating how many LND nodes we are currently connected to (users can trigger alerts based on this)

joostjager:
Do you think we should do a [0,1] metric for each node separately, identified by a label? Given the parallel nature of the node handling, it would require synchronization to bring things together to a single metric.

joostjager:
I suppose it could work with increment and decrement too, if we don't allow any escape routes.

Maybe the metric 'disconnected_node_count' is better for alerting, because then we don't need to know how many nodes we have?

Torakushi:
For me the increment/decrease should do the trick (+ it's thread safe)

I would prefer disconnected_node_count indeed because it is agnostic to our node count.

PR #42, opened 2022-07-26
Initial prometheus metrics support

aakselrod-nydig:
This PR enables initial support of Prometheus monitoring, including basic GRPC client and server support, as specified by #15. Notably, adding custom metrics is out of scope for this PR and will be done as part of a PR addressing #23.

### DIFF
--- a/cmd/lnmuxd/run.go
+++ b/cmd/lnmuxd/run.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"net/http"
 	"os"
 	"os/signal"
 	"syscall"
@@ -17,8 +18,10 @@ import (
 	"github.com/bottlepay/lnmux/persistence"
 	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/go-pg/pg/v10"
+	grpc_prometheus "github.com/grpc-ecosystem/go-grpc-prometheus"
 	"github.com/lightningnetwork/lnd/clock"
 	"github.com/lightningnetwork/lnd/keychain"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/urfave/cli/v2"
 	"golang.org/x/sync/errgroup"
 	"google.golang.org/grpc"
@@ -106,9 +109,13 @@ func runAction(c *cli.Context) error {
 		return err
 	}
 
-	// Instantiate grpc server.
-	grpcServer := grpc.NewServer()
+	// Instantiate grpc server and enable reflection and Prometheus metrics.
+	grpcServer := grpc.NewServer(
+		grpc.StreamInterceptor(grpc_prometheus.StreamServerInterceptor),
+		grpc.UnaryInterceptor(grpc_prometheus.UnaryServerInterceptor),
+	)
 	reflection.Register(grpcServer)
+	grpc_prometheus.Register(grpcServer)
 
 	server := newServer(creator, registry)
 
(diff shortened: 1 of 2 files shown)


cmd/lnmuxd/run.go
```diff
@@ -106,9 +109,10 @@ func runAction(c *cli.Context) error {
 		return err
 	}
 
-	// Instantiate grpc server.
+	// Instantiate grpc server and enable reflection and Prometheus metrics.
 	grpcServer := grpc.NewServer()
 	reflection.Register(grpcServer)
+	grpc_prometheus.Register(grpcServer)
```
joostjager, line 118: Is this all that's needed? I thought there were also interceptors that need to be registered? I also don't think I see the calls to lnmux in the metrics.

For outgoing calls to lnd it may already work. For that, interceptors were added already. (`loadGrpcConn`)

cmd/lnmuxd/run.go
```diff
@@ -126,6 +130,18 @@ func runAction(c *cli.Context) error {
...
+	promAddress := cfg.PrometheusAddress
+	if promAddress == "" {
+		promAddress = DefaultPrometheusAddress
+	}
+
+	// Instantiate a new HTTP server and mux.
+	metrics := &http.Server{
+		Addr:              promAddress,
+		Handler:           promhttp.Handler(),
+		ReadHeaderTimeout: 10 * time.Second,
```
joostjager, line 151: Comment as to why this is necessary?

cmd/lnmuxd/run.go
```diff
@@ -126,6 +130,18 @@ func runAction(c *cli.Context) error {
 		return err
 	}
 
+	promAddress := cfg.PrometheusAddress
+	if promAddress == "" {
+		promAddress = DefaultPrometheusAddress
+	}
+
+	// Instantiate a new HTTP server and mux.
+	metrics := &http.Server{
```
joostjager: Looking ahead at https://github.com/bottlepay/lnmux/issues/43, we may need to use a different http server to allow the registration of multiple endpoints. But can be done later.

cmd/lnmuxd/run.go
```diff
@@ -106,9 +109,10 @@ func runAction(c *cli.Context) error {
 		return err
 	}
 
-	// Instantiate grpc server.
+	// Instantiate grpc server and enable reflection and Prometheus metrics.
```
joostjager, line 112: ~Besides the grpc metrics, what other lnmux-specific metrics do you think are useful to export?~ 

Ok, read in the description that this is for later.

cmd/lnmuxd/run.go
aakselrod-nydig: Sorry, I should have made this into a draft - I'm still working on this.

cmd/lnmuxd/run.go
aakselrod-nydig: I've changed the configuration option to be `InstrumentationAddress` and the related wording. I also changed the mux of the HTTP server to a fresh `ServeMux` and registered the Prometheus handler as `/metrics` so we can just add a `/pprof` handler for #43.

cmd/lnmuxd/run.go
aakselrod-nydig: I added a comment. It's not quite as relevant for a server where inbound is ostensibly only from trusted clients but IMO defense in depth is better than turning off the linter warning.

cmd/lnmuxd/run.go
joostjager: No, my bad. I should have checked whether I was even requested to review!

aakselrod-nydig: I've added interceptors when the server is created, and checked that both the outgoing calls to lnd and incoming calls from an lnmux client are counted in metrics.

cmd/lnmuxd/run.go
joostjager: Ah, there was a linter warning. Interesting.

joostjager (approved):
LGTM

PR #44, opened 2022-07-28
add pprof support

### LINKED ISSUE #43: Add go profiler endpoint
To retrieve things like goroutine and heap profiles.

It is possible to combine this with the prometheus metrics in a single endpoint.

aakselrod-nydig:
This PR allows the HTTP server used for Prometheus metrics to also be used for debugging/profiling using pprof. The handlers are added manually rather than using the default ServeMux. Closes #43.

### DIFF
--- a/cmd/lnmuxd/run.go
+++ b/cmd/lnmuxd/run.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"net/http/pprof"
 	"os"
 	"os/signal"
 	"syscall"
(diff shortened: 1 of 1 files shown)


Torakushi:
I can't believe this problem is still a thing!

especially since it is well known that the default mux has bad perf

Torakushi (approved):
I can't believe this problem is still a thing!

especially since it is well known that the default mux has bad perf

cmd/lnmuxd/run.go
```diff
@@ -149,8 +150,26 @@ func runAction(c *cli.Context) error {
...
+	instMux.HandleFunc("/debug/pprof/cmdline", pprof.Cmdline)
+	instMux.HandleFunc("/debug/pprof/profile", pprof.Profile)
+	instMux.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
+	instMux.HandleFunc("/debug/pprof/trace", pprof.Trace)
+	instMux.Handle("/debug/pprof/allocs", pprof.Handler("allocs"))
+	instMux.Handle("/debug/pprof/block", pprof.Handler("block"))
+	instMux.Handle("/debug/pprof/goroutine", pprof.Handler("goroutine"))
+	instMux.Handle("/debug/pprof/heap", pprof.Handler("heap"))
+	instMux.Handle("/debug/pprof/mutex", pprof.Handler("mutex"))
+	instMux.Handle("/debug/pprof/threadcreate", pprof.Handler("threadcreate"))
```
joostjager, line 171: I've also seen `github.com/go-chi/chi` being used for this, but not sure how it compares.

```
	mux := chi.NewMux()

	mux.HandleFunc("/metrics", handlers.DirectMetricsHandler())
	mux.Mount("/debug", middleware.Profiler())
```

PR #45, opened 2022-07-28
invoiceregistry_test: add test for accept timeout

### LINKED ISSUE #31: Add unit test for accept timeout scenario
There is currently no coverage on that.

aakselrod-nydig:
This adds a test for timing out an invoice after it's accepted but before a settle or cancel is requested. The following increase in test coverage is observed for the `mux` package:

```
ok      github.com/bottlepay/lnmux      2.715s  coverage: 75.7% of statements
```
vs
```
ok      github.com/bottlepay/lnmux      4.763s  coverage: 77.1% of statements
```

This closes #31

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,7 +15,6 @@ jobs:
       postgres:
         image: postgres:14
         env:
-          POSTGRES_DB: bottle_test
           POSTGRES_USER: bottle_test
           POSTGRES_PASSWORD: bottle_test
 
@@ -28,4 +27,4 @@ jobs:
         # populates the client with data, and retrieves data
         run: go test -cover ./...
         env:
-          LNMUX_TEST_DB_DSN: "postgres://bottle_test:bottle_test@postgres/bottle_test?sslmode=disable"
+          LNMUX_TEST_DB_DSN: "postgres://bottle_test:bottle_test@postgres/postgres?sslmode=disable"
--- a/invoiceregistry_test.go
+++ b/invoiceregistry_test.go
@@ -348,7 +348,7 @@ func TestAcceptTimeout(t *testing.T) {
 
 	accept := <-acceptChan
 
-	c.checkHtlcFailed(<-resolved, ResultInvoiceExpired)
+	c.checkHtlcFailed(<-resolved, ResultAcceptTimeout)
 
 	require.ErrorIs(t, c.registry.CancelInvoice(accept.hash, accept.setID), types.ErrInvoiceNotFound)
 }
--- a/invoiceregistry_test.go
+++ b/invoiceregistry_test.go
@@ -10,7 +10,6 @@ import (
 	"github.com/bottlepay/lnmux/test"
 	"github.com/bottlepay/lnmux/types"
 	"github.com/btcsuite/btcd/chaincfg"
-	"github.com/go-pg/pg/v10"
 	"github.com/lightningnetwork/lnd/clock"
 	"github.com/lightningnetwork/lnd/lntypes"
 	"github.com/lightningnetwork/lnd/lnwire"
@@ -22,9 +21,9 @@ import (
 type registryTestContext struct {
 	t               *testing.T
 	registry        *InvoiceRegistry
-	pg              *pg.DB
 	cfg             *RegistryConfig
 	db              *persistence.PostgresPersister
+	dropDB          func()
 	cancelRegistry  func()
 	registryErrChan chan error
 	logger          *zap.SugaredLogger
@@ -37,7 +36,7 @@ type registryTestContext struct {
 func newRegistryTestContext(t *testing.T, autoSettle bool) *registryTestContext {
 	logger, _ := zap.NewDevelopment()
 
-	pg, db := setupTestDB(t)
+	db, dropDB := setupTestDB(t)
 
 	keyRing := NewKeyRing(testKey)
 
@@ -62,9 +61,9 @@ func newRegistryTestContext(t *testing.T, autoSettle bool) *registryTestContext 
 
 	c := &registryTestContext{
 		t:       t,
-		pg:      pg,
 		cfg:     cfg,
 		db:      db,
+		dropDB:  dropDB,
 		logger:  cfg.Logger,
 		testAmt: 10000,
 		creator: creator,
@@ -99,7 +98,7 @@ func (r *registryTestContext) stop() {
 func (r *registryTestContext) close() {
 	r.stop()
 
-	r.pg.Close()
+	r.dropDB()
 }
 
 func (r *registryTestContext) createInvoice(id int, expiry time.Duration) ( // nolint:unparam
@@ -141,6 +140,7 @@ func (r *registryTestContext) checkHtlcSettled(resolution HtlcResolution) {
 
 func TestInvoiceExpiry(t *testing.T) {
 	defer test.Timeout()()
+	t.Parallel()
 
 	c := newRegistryTestContext(t, false)
 
@@ -171,6 +171,7 @@ func TestInvoiceExpiry(t *testing.T) {
 
 func TestAutoSettle(t *testing.T) {
 	defer test.Timeout()()
+	t.Parallel()
 
 	c := newRegistryTestContext(t, true)
 
@@ -207,6 +208,7 @@ func TestAutoSettle(t *testing.T) {
 // application subscribed to accept events.
 func TestNoSubscriberFail(t *testing.T) {
 	defer test.Timeout()()
+	t.Parallel()
 
 	c := newRegistryTestContext(t, false)
 
@@ -233,6 +235,7 @@ func TestNoSubscriberFail(t *testing.T) {
 
 func TestSettle(t *testing.T) {
 	defer test.Timeout()()
+	t.Parallel()
 
 	c := newRegistryTestContext(t, false)
 
@@ -285,6 +288,7 @@ func TestSettle(t *testing.T) {
 
 func TestCancel(t *testing.T) {
 	defer test.Timeout()()
+	t.Parallel()
 
 	c := newRegistryTestContext(t, false)
 
@@ -321,6 +325,7 @@ func TestCancel(t *testing.T) {
 
 func TestAcceptTimeout(t *testing.T) {
 	defer test.Timeout()()
+	t.Parallel()
 
 	c := newRegistryTestContext(t, false)
 
@@ -355,6 +360,7 @@ func TestAcceptTimeout(t *testing.T) {
 
 func TestOverpayment(t *testing.T) {
 	defer test.Timeout()()
+	t.Parallel()
 
 	c := newRegistryTestContext(t, true)
 
--- a/invoiceregistry_test.go
+++ b/invoiceregistry_test.go
@@ -319,6 +319,40 @@ func TestCancel(t *testing.T) {
 	require.ErrorIs(t, c.registry.CancelInvoice(accept.hash, accept.setID), types.ErrInvoiceNotFound)
 }
 
+func TestAcceptTimeout(t *testing.T) {
+	defer test.Timeout()()
+
+	c := newRegistryTestContext(t, false)
+
+	acceptChan, acceptCancel := c.subscribeAccept()
+	defer acceptCancel()
+
+	// Add invoice.
+	invoice, preimage := c.createInvoice(1, time.Hour)
+	hash := preimage.Hash()
+
+	resolved := make(chan HtlcResolution)
+	c.registry.NotifyExitHopHtlc(&registryHtlc{
+		rHash:         hash,
+		amtPaid:       lnwire.MilliSatoshi(c.testAmt),
+		expiry:        100,
+		currentHeight: 0,
+		resolve: func(r HtlcResolution) {
+			resolved <- r
+		},
+		payload: &testPayload{
+			amt:     lnwire.MilliSatoshi(c.testAmt),
+			payAddr: invoice.PaymentAddr,
+		},
+	})
+
+	accept := <-acceptChan
+
+	c.checkHtlcFailed(<-resolved, ResultInvoiceExpired)
+
+	require.ErrorIs(t, c.registry.CancelInvoice(accept.hash, accept.setID), types.ErrInvoiceNotFound)
+}
+
 func TestOverpayment(t *testing.T) {
 	defer test.Timeout()()
 
--- a/persistence/pg_test.go
+++ b/persistence/pg_test.go
@@ -6,27 +6,32 @@ import (
 
 	"github.com/bottlepay/lnmux/persistence/test"
 	"github.com/bottlepay/lnmux/types"
-	"github.com/go-pg/pg/v10"
 	"github.com/lightningnetwork/lnd/lntypes"
 	"github.com/stretchr/testify/require"
 	"go.uber.org/zap"
 )
 
-func setupTestDB(t *testing.T) (*pg.DB, *PostgresPersister) {
-	conn, dsn := test.ResetPGTestDB(t, &test.TestDBSettings{
+func setupTestDB(t *testing.T) (*PostgresPersister, func()) {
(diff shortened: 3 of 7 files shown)


.github/workflows/test.yml
```diff
@@ -26,6 +26,6 @@ jobs:
       - name: Run tests
         # Runs a script that creates a PostgreSQL client,
         # populates the client with data, and retrieves data
-        run: go test -cover ./...
+        run: go test -cover -p 1 ./...
```
aakselrod-nydig: This runs one test at a time to prevent `setupTestDB` from using `ResetPGTestDB` to reset the test database for one test while another is using it. We're up to enough tests now that they interfere with each other in this way. Before this change, only 1/4 test runs succeeded without running into a `Process: ERROR #42P01 relation "lnmux.invoices" does not exist` error, while after this change, 4/4 test runs succeeded.

We may have a few other options on handling this, such as refactoring the tests to explicitly run one at a time from within a test runner that sets up and tears down the database between running tests. However, this doesn't guarantee that tests from one package won't interfere with tests from another package. Another option might to prevent the database resets between tests and instead ensure that each test uses unique invoice IDs, which again may be difficult to coordinate between multiple test files and packages.

This may also extend test run time in the future as we add more tests, which may cause issues with the current test timeout mechanism.

Torakushi: Maybe  you can use different schemas? 
https://github.com/bottlepay/lnmux/blob/75fc8732f1305964c31a09d94f8513d846b8deb6/persistence/test/test.go#L38

Regarding options you mentioned, to ensure that package won't interfere you can create a synchronisation mechanism in `test.go` (lock, ...), and use test suite like [here](https://pkg.go.dev/github.com/maxatome/go-testdeep/helpers/tdsuite). 

Then you can use `Setup` and `Destroy` to lock/unlock the DB. We will stil run packages that use the DB concurrently, but for those who do not use it, we can launch them in parallel.

invoiceregistry_test.go
```diff
@@ -291,6 +291,40 @@ func TestCancel(t *testing.T) {
 	require.ErrorIs(t, c.registry.CancelInvoice(accept.hash, accept.setID), types.ErrInvoiceNotFound)
 }
 
+func TestAcceptTimeout(t *testing.T) {
+	defer test.Timeout()()
+
+	c := newRegistryTestContext(t, false)
```
Torakushi, line 330: IMHO, we need `newRegistryTestContext` to be more configurable.

Indeed I don't like the fact that we need to wait `2* time.second` to reach the accept timeout. [here](https://github.com/aakselrod/lnmux/blob/3fa04d45e1c0a6c1ba20611ca3896c4412246ec1/invoiceregistry_test.go#L37)

If we need to wait 2 seconds for each test, it's going to be very long.

Maybe we can create newRegistryTestContext(RegisterTestContectOption{}) (or whatever)
I am pretty confident that one day we will need to change the `testAmt` as well for example.


I don't know if it has to be done now (lets discuss)

.github/workflows/test.yml
joostjager: Is it right that within a package, we can control this problem by just not marking tests with `t.Parallel()`, but our problem is that both the main package and the persistence package tests execute in parallel?

joostjager: Maybe it is possible to create random databases for each test?

Torakushi: > Is it right that within a package, we can control this problem by just not marking tests with `t.Parallel()`, but our problem is that both the main package and the persistence package tests execute in parallel?

Yeah exactly that, like said the `go help` for the test flag: 
_The 'go test' command may run tests for different packages in parallel as well, according to the setting of the -p flag_


> Maybe it is possible to create random databases for each test?

Yeah we could totally do that but why not using the `schema` mechanism already included in the `test.go`?
The only benefit I see is that, we don't have to check if the schema we want to use for a future test package has already been used by another test package.

invoiceregistry_test.go
```diff
@@ -291,6 +291,40 @@ func TestCancel(t *testing.T) {
...
+		},
+		payload: &testPayload{
+			amt:     lnwire.MilliSatoshi(c.testAmt),
+			payAddr: invoice.PaymentAddr,
+		},
+	})
+
+	accept := <-acceptChan
+
+	c.checkHtlcFailed(<-resolved, ResultInvoiceExpired)
```
joostjager: The test is valid, but I don't think that this failure reason is correct. Ideally we'd distinguish between an invoice expiring and accept timeout.

aakselrod-nydig: I've added a new `ResultAcceptTimeout` to the results and used that in the case that we're failing an invoice for an accept timeout. Also started signing my commits. :)

.github/workflows/test.yml
aakselrod-nydig: I've changed the code to create a new DB for each tests. This means on a local dev box, the user under which the tests runs needs to be able to create databases (`ALTER ROLE bottle WITH CREATEDB;` or similar). I've run the tests several times in CI (e.g. https://github.com/bottlepay/lnmux/actions/runs/2907465477) and the tests no longer seem to flake/run over each other's DB.

There are other choices of how to do this so I'd love some feedback. There may also be opportunities to remove a few more lines from the code if there's consensus that this is the way to move forward for now.

persistence/test/test.go
```diff
@@ -66,5 +78,23 @@ func ResetPGTestDB(t *testing.T, settings *TestDBSettings) (*pg.DB, string) {
 
 	require.Equal(t, settings.ExpectedSchemaVersion, newVersion)
 
-	return db, dsn
+	return db, dbSettings
+}
+
+func DropTestDB(t *testing.T, db *pg.DB) {
+
```
joostjager, line 99: Superfluous newline

persistence/test/test.go
```diff
@@ -45,10 +46,21 @@ func ResetPGTestDB(t *testing.T, settings *TestDBSettings) (*pg.DB, string) {
 
 	dbSettings, err := pg.ParseURL(settings.DSN)
 	require.NoError(t, err)
+
+	defaultDb := pg.Connect(dbSettings)
```
joostjager, line 48: I think we don't necessarily need `bottle_test` anymore now and can always connect to the database with the name `postgres`?

persistence/test/test.go
```diff
@@ -45,10 +46,21 @@ func ResetPGTestDB(t *testing.T, settings *TestDBSettings) (*pg.DB, string) {
...
+	_, err = defaultDb.Exec(fmt.Sprintf("CREATE DATABASE %s", randDbName))
 	require.NoError(t, err)
 
+	dbSettings.Database = randDbName
+
 	db := pg.Connect(dbSettings)
 
+	defaultDb.Close()
+
 	_, err = db.Exec(fmt.Sprintf("DROP SCHEMA IF EXISTS %s CASCADE", settings.Schema))
```
joostjager: No longer necessary to drop schema?

persistence/test/test.go
```diff
@@ -28,12 +29,12 @@ func PGTestDSN() string {
...
-func ResetPGTestDB(t *testing.T, settings *TestDBSettings) (*pg.DB, string) {
+func ResetPGTestDB(t *testing.T, settings *TestDBSettings) (*pg.DB,
+	*pg.Options) {
 
 	if settings.DSN == "" {
 		settings.DSN = PGTestDSN()
 	}
-	dsn := settings.DSN
 
 	if settings.Schema == "" {
```
joostjager: Pre-existing, but is this if statement necessary?

persistence/test/test.go
```diff
@@ -45,10 +46,21 @@ func ResetPGTestDB(t *testing.T, settings *TestDBSettings) (*pg.DB, string) {
 
 	dbSettings, err := pg.ParseURL(settings.DSN)
 	require.NoError(t, err)
+
+	defaultDb := pg.Connect(dbSettings)
+	dbSettings = defaultDb.Options()
+
+	randDbName := fmt.Sprintf("bottle_test_%d", rand.Uint32()) //nolint:gosec
```
joostjager: I wonder if it could be useful to use an (atomic) counter here, so that if a test fails there aren't so many left-over dbs that should have been dropped. Not on CI, but in a local dev env it may get spammy?

persistence/test/test.go
```diff
@@ -66,5 +78,23 @@ func ResetPGTestDB(t *testing.T, settings *TestDBSettings) (*pg.DB, string) {
...
+}
+
+func DropTestDB(t *testing.T, db *pg.DB) {
+
+	opts := db.Options()
+	dbName := db.Options().Database
+
+	db.Close()
+
+	opts.Database = "template1"
```
joostjager: Document that this one is always existing?

persistence/pg_test.go
```diff
@@ -13,20 +13,20 @@ import (
...
+	db := NewPostgresPersisterFromOptions(opts, log)
 
 	return conn, db
 }
 
 func TestSettleInvoice(t *testing.T) {
 	pg, persister := setupTestDB(t)
-	defer pg.Close()
+	defer test.DropTestDB(t, pg)
+	defer persister.Close()
```
joostjager: Perhaps it is somewhat cleaner to return a `close()` function from `setupTestDB` that does both of these things.

persistence/test/test.go
joostjager: Or `template1`, if that's also always present.

joostjager (commented):
Random db approach looks good.

persistence/test/test.go
joostjager: Ok, you already mentioned above that there is potential to remove some lines.

persistence/test/test.go
Torakushi: What happen if there is a collision for DB names? (for example 2 DB having the same name (unfortunately))? 

We will have the same problem as before.

Like Joost said, it could be good to simply have a counter and increment it (atomic).

persistence/test/test.go
aakselrod-nydig: I don't believe `postgres` is always present, but we're now connecting with `template1`. I've updated the GitHub workflow as well.

aakselrod-nydig (commented):
I've fixed up the final commit of the PR. I've removed a lot of extraneous code that isn't needed to support the new model where each test has a brand new DB, added an atomic counter for the DB suffix (with the base still chosen randomly), and a little bit of additional cleanup in the DB test code.

persistence/test/test.go
```diff
@@ -22,39 +26,43 @@ type TestDBSettings struct {
...
-	require.NoError(t, err)
 
-	db := pg.Connect(dbSettings)
+	defaultDb := pg.Connect(dbSettings)
+	dbSettings = defaultDb.Options()
 
-	_, err = db.Exec(fmt.Sprintf("DROP SCHEMA IF EXISTS %s CASCADE", settings.Schema))
-	require.NoError(t, err)
+	randDbName := fmt.Sprintf("bottle_test_%d",
+		atomic.AddUint32(&dbSuffix, 1))
```
joostjager: What is the idea behind starting with a random number and incrementing? Don't we risk left-over dbs on each run still?

persistence/test/test.go
```diff
@@ -22,39 +26,43 @@ type TestDBSettings struct {
...
 	return dsn
 }
 
-func ResetPGTestDB(t *testing.T, settings *TestDBSettings) (*pg.DB, string) {
+func CreatePGTestDB(t *testing.T, settings *TestDBSettings) *pg.Options {
 
+	// To create a DB, we need a connection to the server that isn't to
+	// the nonexistent DB itself. We connect to the template1 DB, which
+	// always exists.
 	if settings.DSN == "" {
```
joostjager: Pre-existing, but isn't `settings.DSN` always empty for our tests?

persistence/test/test.go
aakselrod-nydig: Good catch, thanks! Fixed.

persistence/test/test.go
aakselrod-nydig: Yes. I've removed the DSN and Schema fields from the test struct as well now.

invoiceregistry_test.go
aakselrod-nydig: We're now not passing a schema at all, even optionally. Now we always use the default schema, but create a new DB for each test.

persistence/test/test.go
```diff
@@ -12,49 +13,51 @@ import (
...
+	dbSettings, err := pg.ParseURL(dsn)
 	require.NoError(t, err)
 
-	db := pg.Connect(dbSettings)
+	defaultDb := pg.Connect(dbSettings)
+	dbSettings = defaultDb.Options()
 
-	_, err = db.Exec(fmt.Sprintf("DROP SCHEMA IF EXISTS %s CASCADE", settings.Schema))
-	require.NoError(t, err)
+	randDbName := fmt.Sprintf("bottle_test_%d",
```
joostjager: Rename variable because it is no longer random?

persistence/test/test.go
```diff
@@ -12,49 +13,51 @@ import (
...
+		atomic.AddUint32(&dbSuffix, 1))
 
-	_, err = db.Exec(fmt.Sprintf("CREATE SCHEMA %s", settings.Schema))
+	_, err = defaultDb.Exec(fmt.Sprintf("CREATE DATABASE %s", randDbName))
```
joostjager: What happens here if the db already exists because of a failed previous run?

invoiceregistry_test.go
Torakushi: I still don't like the fact that we need to wait 2 second for this test.
 And more generally to have a registryContext not configurable
 
Do you think it is overkill to make it configurable? ( Like I said, one day we might need to change the testAmt)

Torakushi: I still don't like the fact that we need to wait 2 second for this test.
And more generally to have a registryContext not configurable

Do you think it is overkill to make it configurable? ( Like I said, one day we might need to change the testAmt)

persistence/test/test.go
aakselrod-nydig: Fixed.

persistence/test/test.go
aakselrod-nydig: Generally, failed runs clean up after themselves (`t.Cleanup()` calls `c.close()` which drops the test DB), but when I created a DB manually and set its owner to the role used by the tests, the code complained that the DB already existed on `CREATE DATABASE`.

I've added a `DROP DATABASE IF EXISTS` above this, so the new code drops the DB and moves on.

invoiceregistry_test.go
aakselrod-nydig: I've parallelized the tests now, so the 2 seconds per test wouldn't be cumulative but in parallel. I do like the idea of making registryTestContext configurable, so I'll take a quick stab at it.

Torakushi: It could be for a next PR, IMHO it could be useful in the future :)

aakselrod-nydig: Sounds good, I've added #60 for a future PR.

persistence/test/test.go
```diff
@@ -66,5 +80,24 @@ func ResetPGTestDB(t *testing.T, settings *TestDBSettings) (*pg.DB, string) {
...
+
+func DropTestDB(t *testing.T, opts pg.Options) {
+	dbName := opts.Database
+
+	// To drop the DB, we need a connection to the server that isn't
+	// accessing the DB itself. We connect to the template1 DB, which
+	// always exists.
+	opts.Database = "template1"
+
+	template1 := pg.Connect(&opts)
```
joostjager: Doesn't this need the same lock again?

It does make me wonder if connecting to `template1` is such a good idea.

Are you sure `postgres` isn't also available generally? That is what people on the internet seem to recommend.

aakselrod-nydig: Per discussion, we removed the lock altogether, replaced `template1` with `postgres` (which should work in most situations), and added some context from the caller package to also remove conflicts between tests from different packages running in separate binaries.

joostjager (approved):

PR #46, opened 2022-08-01
interceptor: handle error channel on initial block height

aakselrod-nydig:
This PR handles the case where the interceptor gets an error when subscribing to block epoch notification. This can help surface problems such as `lnd` built without the `chainrpc` tag.


Torakushi (approved):
LGTM

PR #48, opened 2022-08-03
config.go: do not return error if there are unknown fields

Torakushi:
Integrating lnmux in a production environment may require adding fields to the configuration ((We can imagine a system where the config needs to have specific fields to be accepted...).

However, it is currently not possible because we use UnmarshalStrict. We suggest using Unmarshal instead.


aakselrod-nydig:
LGTM

PR #51, opened 2022-08-05
config.go: possibility to use env variables to override dsn and identity key

### LINKED ISSUE #50: Environment variables to override some settings in the configuration
Currently we use a `yaml` file for lnmux configuration (nothing more, nothing less).
We will need an other mechanism to override/set some settings in the configuration in order to: 

-   protect the `identity key`
-   avoid putting credential for the configuration (for the db)

At first, we considered using `viper` (for the env variable mechanism). But LND stopped using it, so we don't want to add a dependency on it.

We propose to use two different environment variables:

-  `LNMUX_PERSISTENCE_POSTGRES_DSN` for the postgres dsn
-  `LNMUX_IDENTITY_KEY` for the identity key

Torakushi:
This adds two env variables: `LNMUX_PERSISTENCE_POSTGRES_DSN` (for the postgres dsn) and
`LNMUX_IDENTITY_KEY` (for the identity key) in order to set/override settings in the configuration

We find it relevant to use environment variables only for DSN and identity key to improve security and avoid exposing secret.
If we want to extend this behaviour for other config settings (without `viper` or other dependencies), a proposition can be found [bellow](https://github.com/bottlepay/lnmux/pull/51/files#r938704264)

PS: This replaces `ioutils` by `os` because `ioutils` is DEPRECIATED.
PS2: use `const` instead of `var` when possible

close #50


cmd/lnmuxd/config.go
```diff
@@ -117,5 +120,18 @@ func loadConfig(filename string) (*Config, error) {
...
+	// Check if the user provides an environment variable for PG DSN and load it
+	loadStringEnvVariable(PostgresDSNEnvKey, &cfg.DB.DSN)
+
+	// Check if the user provides an environment variable for the identity key and load it
+	loadStringEnvVariable(IdentityKeyEnvKey, &cfg.IdentityKey)
+
 	return &cfg, nil
 }
+
+func loadStringEnvVariable(key string, value *string) {
```
Torakushi, line 132: If we need to extend this behavior we can either create X functions for each type or use generics/interface.

For example:

```
func loadEnvVariableGen[T any](key string, value T) {
	if val, ok := os.LookupEnv(key); ok {
		log.Infof("Environment variable %q found", key)
		if reflect.ValueOf(value).Kind() != reflect.Ptr {
			log.Info("A pointer is needed for 'loadEnvVariableGen', environment variable discarded")
			return
		}

		switch t := any(value).(type) {
		case *string:
			*(any(value).(*string)) = val
                default
                	log.Infof("Don't know how to load environment variable in type %T", t)
		}
	}
}
```

keyring.go
```diff
@@ -70,7 +70,7 @@ func (b *KeyRing) DerivePrivKey(keyDesc keychain.KeyDescriptor) (
 // k is our private key, and P is the public key, we perform the following
 // operation:
 //
-//  sx := k*P s := sha256(sx.SerializeCompressed())
+// sx := k*P s := sha256(sx.SerializeCompressed())
```
Torakushi, line 73: I had to do that because now `golangci/golangci-lint-action@v3` use the new version of golangci-lint (1.48.0).

This double space was detected as an error 

NOW:                `Requested golangci-lint 'latest', using 'v1.48.0'`
BEFORE (yesterday):          `Requested golangci-lint 'latest', using 'v1.47.3'`

README.md
```diff
@@ -13,14 +13,14 @@ For the setup, it is assumed that there are multiple LND nodes running with conn
...
 
 * Create a postgres database. Note that `lnmux` uses stateless invoices. This means that the database only contains settled invoices.
 
 * Create a config file for `lnmuxd` named `lnmux.yml`. An [example](lnmux.yml.example) can be found in this repository. The config file contains the following elements:
   * LND nodes configuration: TLS certificate, macaroon, address and pubkey. Pubkey is configured as a protection against unintentionally connecting to the wrong node.
-  * Postgres connection string for the database created above.
-  * 32-byte Identity PRIVATE key. This key is used to generate invoices and decode incoming htlcs. In a production environment, this key must be protected carefully.
+  * Postgres connection string for the database created above (can be overridden by the environment variable `LNMUX_PERSISTENCE_POSTGRES_DSN` to avoid exposing credentials).
+  * 32-byte Identity PRIVATE key. This key is used to generate invoices and decode incoming htlcs. In a production environment, this key must be protected carefully (can be overridden by the environment variable `LNMUX_IDENTITY_KEY` to avoid exposing the private key).
```
aakselrod-nydig, line 23: 🔥 Thanks for updating the docs!

keyring.go
aakselrod-nydig: Good catch, thank you!

cmd/lnmuxd/config.go
aakselrod-nydig: I like the simple version for now. Looks great!

aakselrod-nydig (approved):

PR #53, opened 2022-08-19
don't add description in 'zpay32.Invoice' if it is an empty string

### LINKED ISSUE #52: AddInvoice returns error "both description and description hash set" even if description is empty
When creating an invoice with lnmux (using zpay32), an error is returned if we set description and descriptionHash ([here](https://github.com/lightningnetwork/lnd/blob/master/zpay32/invoice.go#L360))

The field `description` for `AddInvoiceRequest` is a string so, even if we put am empty string in `description`, after doing `zpay32.Description(description)`, we will have a `zpay32.Invoice.Description` not nil

(because zpay32.Description(description) will take the address of the empty string `description`)

A simple fix would be to check if `description == ""` before doing `zpay32.Description(description)`

Torakushi:
close #52

### DIFF
--- a/invoice_creator.go
+++ b/invoice_creator.go
@@ -137,9 +137,12 @@ func (c *InvoiceCreator) Create(amtMSat int64, expiry time.Duration,
 
 	options = append(options, zpay32.Expiry(expiry))
 
-	// Use the memo field as the description. If this is not set
-	// this will just be an empty string.
-	options = append(options, zpay32.Description(memo))
+	// Use the memo field as the description if it's not empty.
+	// Otherwise, we could have the error "both description and description hash set" in zpay32.NewInvoice
+	// even if `memo` is empty.
+	if len(memo) > 0 {
+		options = append(options, zpay32.Description(memo))
+	}
 
 	if descHash != nil {
 		options = append(options, zpay32.DescriptionHash(*descHash))


invoice_creator.go
```diff
@@ -137,9 +137,12 @@ func (c *InvoiceCreator) Create(amtMSat int64, expiry time.Duration,
...
-	// this will just be an empty string.
-	options = append(options, zpay32.Description(memo))
+	// Use the memo field as the description if it's not empty.
+	// Otherwise, we could have the error "both description and description hash set" in zpay32.NewInvoice
+	// even if `memo` is empty.
+	if len(memo) > 0 {
+		options = append(options, zpay32.Description(memo))
+	}
 
 	if descHash != nil {
```
aakselrod-nydig, line 147: Should this line also change? If descHash is provided but memo is also >0, then I think we still end up with both. Instead, I think this line should be:

`if descHash != nil && len(memo) == 0 {`

Generally the caller shouldn't set descHash in this case, but it would be good to be sure?

aakselrod-nydig (changes requested):
This looks great, just one small suggestion.

invoice_creator.go
Torakushi: Actually no. We had a conversation with @joostjager and ignoring one of the two field is not a good idea. (even if it is done that way in LND [here](https://github.com/lightningnetwork/lnd/blob/master/lnrpc/invoicesrpc/addinvoice.go#L322)).

It should be the responsibility of the application (that uses lnmux) to ensure that it does not give a description and a descriptionhash at the same time

aakselrod-nydig: Sounds good, I'll merge now.

aakselrod-nydig (approved):
Per internal discussion, better to leave the responsibility of determining memo vs. descHash fully to the caller.

PR #56, opened 2022-08-22
invoice without description should be allowed

### LINKED ISSUE #55: Invoices without 'Description' should be allowed
When creating an invoice, if 'Description' is empty and 'DescriptionHash' is nil, `zpay32` returns an error:

`neither description nor description hash set`

Invoice should be allowed even without 'Description'.

In that case, simply add an empty string for Description  (`options = append(options, zpay32.Description(""))` )

Torakushi:
close #55

### DIFF
--- a/invoice_creator.go
+++ b/invoice_creator.go
@@ -140,7 +140,9 @@ func (c *InvoiceCreator) Create(amtMSat int64, expiry time.Duration,
 	// Use the memo field as the description if it's not empty.
 	// Otherwise, we could have the error "both description and description hash set" in zpay32.NewInvoice
 	// even if `memo` is empty.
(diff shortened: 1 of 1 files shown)


joostjager (approved):

invoice_creator.go
```diff
@@ -137,17 +137,19 @@ func (c *InvoiceCreator) Create(amtMSat int64, expiry time.Duration,
 
 	options = append(options, zpay32.Expiry(expiry))
 
+	if descHash != nil {
```
joostjager: Moving this up wasn't really necessary I think?

Torakushi: No indeed, I changed my way of doing things along the way.

it's a vestige

PR #57, opened 2022-08-23
distinct virtual channel id per gateway node

### LINKED ISSUE #36: Unique virtual channel id
Currently the virtual channel that routes to lnmux is always 12345 for every lnd instance. It is probably safer to change this to a more realistic value that is also unique per instance. Less leakage of implementation details and also we rule out the possibility of some implementation tripping over the duplicate channel id in the route hints.

LDK for inspiration: https://github.com/lightningdevkit/rust-lightning/blob/f26e8541a516edd4d6493cf2aa8d700ba6bbd7ca/lightning/src/util/scid_utils.rs#L65

joostjager:
Simple change to have distinct virtual channel ids. Randomizing seemed to be unnecessary and may lead to bloating of the reputation database depending on the sender's node implementation. This does not apply to lnd, because lnd tracks reputation on a per-node basis rather than per-channel.

Also if a sender node tracks performance per-channel and one of the destination nodes is down, they will keep running into the same failure for subsequent payments when using random ids.

fixes #36

### DIFF
--- a/invoice_creator.go
+++ b/invoice_creator.go
@@ -18,7 +18,20 @@ import (
 
 var byteOrder = binary.BigEndian
 
-const virtualChannel = 12345
+// virtualChannelFromNode deterministically derives a short channel id for
+// multiplexer for a node pubkey.
+func virtualChannelFromNode(key common.PubKey) uint64 {
+	var b [8]byte
+
+	// Take the final bytes of the key.
+	copy(b[:], key[common.PubKeySize-8:])
+
+	// Force the high bit to make sure there are no collisions with real
+	// channels.
+	b[0] |= 1 << 7
+
+	return byteOrder.Uint64(b[:])
+}
 
 type InvoiceCreatorConfig struct {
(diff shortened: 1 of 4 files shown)


Torakushi (approved):
LGTM

invoice_creator.go
```diff
@@ -18,7 +18,20 @@ import (
...
 var byteOrder = binary.BigEndian
 
-const virtualChannel = 12345
+// virtualChannelFromNode deterministically derives a short channel id for
+// multiplexer for a node pubkey.
+func virtualChannelFromNode(key common.PubKey) uint64 {
+	var b [8]byte
+
+	// Take the final bytes of the key.
+	copy(b[:], key[33-8:])
```
aakselrod-nydig: Would it be better to use `PubKeySize` instead of `33` here?

aakselrod-nydig (approved):
LGTM.

invoice_creator.go
joostjager: Yes, fixed.

PR #58, opened 2022-08-23
fail htlc without subscriber

### LINKED ISSUE #32: Reject htlcs if there is no accept subscriber
If the application(s) connecting to lnmux are down, it makes no sense to attempt the collection of an htlc set.

joostjager:
fixes #32


Torakushi (approved):
Good one to add this logic only if we start a new set

aakselrod-nydig (approved):
LGTM, I like this approach.

PR #59, opened 2022-08-24
remove subscriber requirement in auto-settle mode

joostjager:
Otherwise lnmux is still not auto-settling.

### DIFF
--- a/invoiceregistry.go
+++ b/invoiceregistry.go
@@ -759,7 +759,9 @@ func (i *InvoiceRegistry) process(ctx context.Context, h *registryHtlc) error {
 		//
 		// When we are adding to an existing set, we keep going and assume that
 		// the application will be back online soon.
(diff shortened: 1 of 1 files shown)


aakselrod-nydig (approved):
This makes sense to me. Would be good to add a test for this behavior (maybe inside `AutoSettle`) eventually. Also with auto-settling without subscriber, it would be good to eventually add RPCs for catching up on invoices (maybe a `ListInvoice` with a field to specify the last invoice the subscriber has seen settle)

Torakushi (approved):
If tests are difficult to do, LGTM

joostjager:
Yes, I think we have to decide what we want to do with auto-settle going forward. It's not clear to me if this is a desired feature at all. So far, it has been mainly useful during development to skip a few rpc calls.

And indeed, as @aakselrod-nydig mentions, how can you get an overview of all settled invoices.

PR #63, opened 2022-08-24
Add configuration options for newRegistryTestContext

### LINKED ISSUE #60: Make `registryTestContext` configurable
We are entering the age of complexity with `registryTestContext`. We already pass an `autoSettle bool` option to the `newRegistryTestContext` function, and soon we will need it for `testAmt` and different timeouts.

We could add functional option arguments to `newRegistryTestContext` that would modify the default configuration with the desired option. This would probably add the fewest lines per test, as our defaults should be good for most tests.

aakselrod-nydig:
This PR adds functional option parameter support to `newRegistryTestContext`. It adds several options that would be useful to be able to change within tests, including converting the `autoSettle` argument to an option. It also modifies the `TestAutoSettle` test to use all of the options to show that the auto-settling happens without the accept timer or the HTLC hold timer being taken into account.

Fixes #60

### DIFF
--- a/invoiceregistry_test.go
+++ b/invoiceregistry_test.go
@@ -18,6 +18,8 @@ import (
 	"go.uber.org/zap"
 )
 
+// registryTestContext contains all the elements needed to spin up a registry
+// for testing.
 type registryTestContext struct {
 	t               *testing.T
 	registry        *InvoiceRegistry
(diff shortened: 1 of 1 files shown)


invoiceregistry_test.go
```diff
@@ -19,6 +19,8 @@ import (
 	"go.uber.org/zap"
 )
 
+// registryTestContext contains all the elements needed to spin up a registry
+// for testing..
```
joostjager: double dot

invoiceregistry_test.go
```diff
@@ -172,7 +220,14 @@ func TestInvoiceExpiry(t *testing.T) {
...
 	defer test.Timeout()()
 
-	c := newRegistryTestContext(t, true)
+	c := newRegistryTestContext(t,
+		optAutoSettle(true),
+		optTestAmt(1000),
+
+		// Immediate timeout.
+		optAcceptTimeout(0),
+		optHtlcHoldDuration(0),
```
joostjager: Aren't these zero timeouts potentially causing race conditions?

joostjager (commented):
Don't you want to reduce the timeout on the accept timeout test, the part that started this conversation?

invoiceregistry_test.go
aakselrod-nydig: Fixed, thanks!

invoiceregistry_test.go
aakselrod-nydig: I've taken them out since they aren't super relevant to this test, but I don't believe so.

Auto-settle happens here, in `process()`:
https://github.com/bottlepay/lnmux/blob/master/invoiceregistry.go#L813

`process()` is called from `invoiceEventLoop` when an HTLC comes in here:
https://github.com/bottlepay/lnmux/blob/master/invoiceregistry.go#L297

This happens inside the same goroutine and in the same loop that handles the timeouts and the HTLC accepts here:
https://github.com/bottlepay/lnmux/blob/master/invoiceregistry.go#L277

I believe that means that the HTLCs get settled from lnmux's view before either of the timers would expire, even in an MPP situation where it's an HTLC that completes the set.

I did run the test (without any others at the same time) in this form with the `-race` flag and didn't see it trigger a race condition. However, after I changed the code back to initialize this test context with only autosettle, and ran the entire test suite with the `-race` flag, I noticed there were still races in `TestCancel` and `TestAutoSettle`. I've created #64 to address it.

aakselrod-nydig:
> Don't you want to reduce the timeout on the accept timeout test, the part that started this conversation?

Sorry, I hadn't yet gotten that part in this PR because I based the PR initially on `master` before #45 went in. I've rebased on post-45 `master`, updated the code to remove the modifications to `TestAutoSettle` and made a modification to the newly-included `TestAcceptTimeout`. I also added `// nolint:unused` to a couple of the options we're not currently using.

invoiceregistry_test.go
joostjager: Indeed, I think you are right. The timeout channels don't even get to be evaluated before the settle happens.

invoiceregistry_test.go
```diff
@@ -33,7 +35,48 @@ type registryTestContext struct {
...
+// that modifies the AutoSettle configuration in the RegistryConfig.
+func optAutoSettle(autoSettle bool) regTestContextOption {
+	return func(c *registryTestContext) {
+		c.cfg.AutoSettle = autoSettle
+	}
+}
+
+// optHtlcHoldDuration returns a functional argument to newRegistryTestContext()
+// that modifies the HtlcHoldDuration configuration in the RegistryConfig.
+func optHtlcHoldDuration(d time.Duration) regTestContextOption { // nolint:unused
```
joostjager, line 52: For the next time, I'd opt for not adding dead code in the first place.

joostjager (approved):

Torakushi (approved):
LGTM, Ty for doing this

PR #65, opened 2022-08-25
Enable race detection in CI, fix a data race

### LINKED ISSUE #64: Run tests in CI with race detection enabled
We may have some data races, so we should run our tests to be able to detect them automatically so we can fix them. This should include fixing any races detected after first enabling it.

aakselrod-nydig:
This PR enables race detection when running tests in CI. It also fixes a data race in the database migration discovery process, found automatically when running tests.

Fixes #64

### DIFF
--- a/persistence/migrations/migrations.go
+++ b/persistence/migrations/migrations.go
@@ -1,11 +1,16 @@
 package migrations
 
 import (
+	"sync"
+
 	"github.com/go-pg/migrations/v8"
 )
 
-var Collection = migrations.NewCollection()
-var sqlDiscovered = false
+var (
+	sqlDiscoveryOnce sync.Once
+
+	Collection = migrations.NewCollection()
+)
 
 func init() {
 	Collection.DisableSQLAutodiscover(true)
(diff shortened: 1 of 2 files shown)


persistence/migrations/migrations.go
```diff
@@ -15,9 +19,17 @@ func init() {
 
 // DiscoverSQLMigrations scans dir for migrations.
 func DiscoverSQLMigrations(dir string) error {
+	sqlDiscoveryMtx.Lock()
+	defer sqlDiscoveryMtx.Unlock()
+
+	if sqlDiscovered {
+		return nil
```
joostjager: I think this can be done in a more compact way using `sync.Once`

persistence/migrations/migrations.go
```diff
@@ -1,9 +1,13 @@
...
 
 import (
+	"sync"
+
 	"github.com/go-pg/migrations/v8"
 )
 
+var sqlDiscoveryMtx sync.Mutex
+
 var Collection = migrations.NewCollection()
```
Torakushi: That's nothing but use
var (

)

To group these declarations

persistence/migrations/migrations.go
aakselrod-nydig: I've updated it using `sync.Once`, it's quite a bit more compact. Thanks!

persistence/migrations/migrations.go
aakselrod-nydig: I've added the grouping syntax, but reduced the number of variables here back to 2. Should I still group, and should I keep an empty line between the two variable declarations?

Torakushi: Looks good (that was only a little note!)

Torakushi (approved):
LGTM

joostjager (approved):

PR #66, opened 2022-08-25
Add tests in `InvoiceRegistry` for MPP validity checks

### LINKED ISSUE #62: Add more test coverage for InvoiceRegistry
Coverage analysis shows that we are a bit lacking in some of the failure cases of `invoiceRegistry.process`. Invalid sets, missing mpp fields and the likes.

aakselrod-nydig:
This PR adds a test for some MPP validity checks in the `InvoiceRegistry`'s `process()` function. It currently increases code coverage for `invoiceregistry.go` from 78.9% to 83.3%, and coverage for the `process()` function specifically from 76.8% to 92.8%.

It also:
- fixes the return code when invoice and HTLC payment hash don't match
- adds a set of test utility functions for crafting and sending HTLCs to the registry
- adds a null MPP record type for tests

Fixes #62.

### DIFF
--- a/invoiceregistry.go
+++ b/invoiceregistry.go
@@ -679,7 +679,7 @@ func (i *InvoiceRegistry) process(ctx context.Context, h *registryHtlc) error {
 	if statelessData.preimage.Hash() != h.rHash {
 		logger.Debugw("Hash mismatch")
 
-		h.resolve(NewFailResolution(ResultInvoiceNotFound))
+		h.resolve(NewFailResolution(ResultAddressMismatch))
 
 		return nil
 	}
--- a/invoiceregistry_test.go
+++ b/invoiceregistry_test.go
@@ -28,6 +28,7 @@ type registryTestContext struct {
 	dropDB          func()
 	cancelRegistry  func()
 	registryErrChan chan error
+	resolved        chan HtlcResolution
 	logger          *zap.SugaredLogger
 
 	testAmt int64
@@ -125,6 +126,8 @@ func newRegistryTestContext(t *testing.T,
 }
 
 func (r *registryTestContext) start() {
+	r.resolved = make(chan HtlcResolution)
+
 	r.registryErrChan = make(chan error)
 	r.registry = NewRegistry(r.db, r.cfg)
 
@@ -144,6 +147,12 @@ func (r *registryTestContext) stop() {
 }
 
 func (r *registryTestContext) close() {
+	select {
+	case <-r.resolved:
+		require.FailNow(r.t, "Resolution generated incorrectly")
+	default:
+	}
+
 	r.stop()
 
 	r.dropDB()
@@ -175,15 +184,60 @@ func (r *registryTestContext) subscribeAccept() (chan acceptEvent, func()) {
 	return updateChan, cancel
 }
 
-func (r *registryTestContext) checkHtlcFailed(resolution HtlcResolution,
-	result FailResolutionResult) {
+func (r *registryTestContext) checkHtlcFailed(result FailResolutionResult) {
+	resolution := <-r.resolved
 
 	require.IsType(r.t, &HtlcFailResolution{}, resolution)
 	require.Equal(r.t, result, resolution.(*HtlcFailResolution).Outcome)
 }
 
-func (r *registryTestContext) checkHtlcSettled(resolution HtlcResolution) {
-	require.IsType(r.t, &HtlcSettleResolution{}, resolution)
+func (r *registryTestContext) checkHtlcSettled() {
+	require.IsType(r.t, &HtlcSettleResolution{}, <-r.resolved)
+}
+
+// optNotifyHtlc is a functional option to notifyHtlc.
+type optNotifyHtlc func(h *registryHtlc)
+
+// optAmtPaid changes the amtPaid field of the HTLC.
+func optAmtPaid(amtPaid int64) optNotifyHtlc {
+	return func(h *registryHtlc) {
+		h.amtPaid = msat(amtPaid)
+	}
+}
+
+// optHtlcID sets the HTLC ID for the HTLC's circuit key.
+func optHtlcID(id uint64) optNotifyHtlc {
+	return func(h *registryHtlc) {
+		h.circuitKey.HtlcID = id
+	}
+}
+
+// notifyHtlc notifies the registry of an incoming HTLC. It builds the HTLC
+// with convenient defaults.
+func (r *registryTestContext) notifyHtlc(rHash lntypes.Hash, payAddr [32]byte,
+	opts ...optNotifyHtlc) {
+
+	h := &registryHtlc{
+		rHash:         rHash,
+		amtPaid:       msat(r.testAmt),
+		expiry:        100,
+		currentHeight: 0,
+		payload: &testPayload{
+			amt:     msat(r.testAmt),
+			payAddr: payAddr,
+		},
+		resolve: func(hr HtlcResolution) {
+			go func() {
+				r.resolved <- hr
+			}()
+		},
+	}
+
+	for _, opt := range opts {
+		opt(h)
+	}
+
+	r.registry.NotifyExitHopHtlc(h)
 }
 
 func TestInvoiceExpiry(t *testing.T) {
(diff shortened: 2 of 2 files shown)


invoiceregistry_test.go
```diff
@@ -496,11 +496,93 @@ func TestOverpayment(t *testing.T) {
...
+				HtlcID: id,
+			},
+			resolve: func(r HtlcResolution) {
+				resolved <- r
+			},
+			payload: payload,
+		})
+	}
+
+	// Create invoice that's expired.
```
joostjager: Comment incorrect?

invoiceregistry_test.go
```diff
@@ -496,11 +496,93 @@ func TestOverpayment(t *testing.T) {
...
+
 type testPayload struct {
 	amt     lnwire.MilliSatoshi
 	payAddr [32]byte
+
+	nullMpp bool
 }
 
 func (t *testPayload) MultiPath() *record.MPP {
+	if t.nullMpp {
```
joostjager: You may consider creating two test structs here, to avoid the if and perhaps make it slightly more clear.

invoiceregistry_test.go
```diff
@@ -496,11 +496,93 @@ func TestOverpayment(t *testing.T) {
...
+
+	c.checkHtlcFailed(<-resolved, ResultHtlcSetTotalTooLow)
+
+	// Test MPP with hash mismatching the invoice.
+	notifyHtlc(3, lntypes.ZeroHash, testAmt, 0, &testPayload{
+		amt:     testAmt,
+		payAddr: invoice.PaymentAddr,
+	})
+
+	// TODO(aakselrod): should this be ResultAddressMismatch instead?
```
joostjager: Yes, I think it makes sense to differentiate here.

invoiceregistry_test.go
```diff
@@ -496,11 +496,93 @@ func TestOverpayment(t *testing.T) {
...
 
+func TestMPPValidity(t *testing.T) {
+	defer test.Timeout()()
+	t.Parallel()
+
+	c := newRegistryTestContext(t, optAutoSettle(true))
+
+	resolved := make(chan HtlcResolution)
+
+	notifyHtlc := func(id uint64, rHash lntypes.Hash,
```
joostjager: This one is repeated a lot. Maybe for the next time, we can see if it can be extracted to a method on the context. Would save quite a few lines in this file.

joostjager (commented):
NIce addition, all gaps covered now it seems

invoiceregistry_test.go
aakselrod-nydig: Fixed, thanks!

invoiceregistry_test.go
aakselrod-nydig: I've done it, hopefully it's a bit more clear.

invoiceregistry_test.go
aakselrod-nydig: Fixed!

invoiceregistry_test.go
aakselrod-nydig: I've factored out this inner function into a method on the context, set convenient defaults, and added optional arguments for use in tests similar to what we did with starting a registry context. I've also changed all of the tests to now use this method. It added a bit of setup but now there's far less boilerplate in each existing test (see [here]( https://github.com/bottlepay/lnmux/compare/e1fd6a1f2cd71b38c4e12a9eeb6803795662db09..b94242ebdcd74a6619a5371ccc8bad0334a29b8d#diff-2d9b3b96fa8f72553c7f4c6690edbd5282dd7dfbaf62476d630f5469df8f519f) - 146 additions, 153 deletions) and future tests.

invoiceregistry_test.go
```diff
@@ -186,6 +186,100 @@ func (r *registryTestContext) checkHtlcSettled(resolution HtlcResolution) {
...
+
+// optCurrentHeight sets the HTLC's current block height.
+func optCurrentHeight(height int32) optNotifyHtlc {
+	return func(h *registryHtlc) {
+		h.currentHeight = height
+	}
+}
+
+// notifyHtlc notifies the registry of an incoming HTLC. It builds the HTLC
+// with convenient defaults but requires a minimum of an optRHash and an
```
joostjager: If those two are minimally required, maybe cleaner to make them explicit arguments of `notifyHtlc`?

invoiceregistry_test.go
```diff
@@ -186,6 +186,100 @@ func (r *registryTestContext) checkHtlcSettled(resolution HtlcResolution) {
 	require.IsType(r.t, &HtlcSettleResolution{}, resolution)
 }
 
+// optNotifyHtlc is a functional option to notifyHtlc.
+type optNotifyHtlc func(h *registryHtlc)
```
joostjager, line 199: The `opt` thing is quite useful again to keep tests compact!

invoiceregistry.go
```diff
@@ -679,7 +679,7 @@ func (i *InvoiceRegistry) process(ctx context.Context, h *registryHtlc) error {
 	if statelessData.preimage.Hash() != h.rHash {
 		logger.Debugw("Hash mismatch")
 
-		h.resolve(NewFailResolution(ResultInvoiceNotFound))
+		h.resolve(NewFailResolution(ResultAddressMismatch))
```
joostjager, line 682: The commit message lists a set of changes, which is already an indication that they probably could have been separate commits. I think three:
1. Fix return code
2. Refactor tests / add util func
3. Add more mpp coverage

invoiceregistry_test.go
```diff
@@ -225,27 +308,17 @@ func TestAutoSettle(t *testing.T) {
...
-		resolve: func(r HtlcResolution) {
-			close(resolved)
-		},
-		payload: &testPayload{
-			amt:     lnwire.MilliSatoshi(c.testAmt),
-			payAddr: invoice.PaymentAddr,
-		},
-	})
+	c.notifyHtlc(optRHash(hash), optPayAddr(invoice.PaymentAddr),
+		optResolveClose(resolved))
```
joostjager: Looking at these tests in their cleaned-up form, it seems that the distinction between send and close isn't that useful. Instead of close, a normal send and waiting for the send would probably work just as well.

With that change, you could drop both `optResolveClose` and `optResolveSend` and make the `resolved` channel a property of the context too.

invoiceregistry_test.go
joostjager: Yes, better

invoiceregistry_test.go
aakselrod-nydig: Done. Since there's not a `nil` for the `[32]byte` type of the `payAddr`, I've also added a `payAddrZero` var to compare/assign when intentionally attempting to create an HTLC without an MPP payload.

invoiceregistry.go
aakselrod-nydig: Done.

invoiceregistry_test.go
aakselrod-nydig: Done. Big change in readability and compactness. Also allowed me to change most of the `<-resolved` reads into checks for failure/settlement, except where we're looking for no read on the channels in `TestOverpayment`.

invoiceregistry_test.go
```diff
@@ -464,18 +417,21 @@ func TestOverpayment(t *testing.T) {
...
+	c.checkHtlcFailed(ResultHtlcSetOverpayment)
 
 	// We should not have a second resolution, an update, or an accept here.
 	select {
 
 	case <-acceptChan:
 		require.FailNow(t, "Invoice accepted incorrectly")
 
-	case <-resolved:
+	case <-c.resolved:
```
joostjager: Side remark, not necessarily to be solved in this PR, but this check for unexpected resolutions could be made a standard part of `c.close()`, so that it is checked everywhere.

invoiceregistry_test.go
joostjager: Doesn't that mean that `payAddr` isn't a required parameter after all?

invoiceregistry_test.go
joostjager: Very nice!

invoiceregistry_test.go
aakselrod-nydig: It's required by `process`, but that requirement is part of what we test in the new `TestMPPValidity`.

aakselrod-nydig: I removed the `payAddrZero` and check, and instead created an `optNullPayload` that sets the payload to `(*nullPayload)(nil)` in the option instead. Seems cleaner.

invoiceregistry_test.go
aakselrod-nydig: I made the change in this PR. It doesn't check the `acceptChan` any more, but I think that was somewhat redundant.

invoiceregistry_test.go
joostjager: Agreed

invoiceregistry_test.go
joostjager: More clean up, good

joostjager (approved):
Nice work

PR #68, opened 2022-09-01
WIP: ensure only one instance of lnmuxd can run at a time

aakselrod-nydig:


persistence/pg.go
```diff
@@ -175,7 +175,7 @@ func (p *PostgresPersister) MarkHtlcSettled(ctx context.Context,
 			HtlcID: key.HtlcID,
 		}
 
-		result, err := p.conn.ModelContext(ctx, &htlc).
+		result, err := tx.ModelContext(ctx, &htlc).
```
joostjager, line 262: This is bad. I think you can make a separate PR for this fix.

persistence/migrations/1_initial.up.sql
```diff
@@ -26,3 +26,10 @@ CREATE TABLE htlcs
 
         CONSTRAINT fk_hash FOREIGN KEY(hash) REFERENCES invoices(hash) ON DELETE CASCADE
 );
+
+CREATE TABLE instance_lock
+(
+	"lock_updated_at" TIMESTAMPTZ
+);
+
+INSERT INTO instance_lock VALUES(NOW() - INTERVAL '1 minute');
```
joostjager, line 35: Would squash this commit with the final one that uses the table, to give it more context.

persistence/pg.go
```diff
@@ -58,30 +58,75 @@ type dbHtlc struct {
...
+	// timestamp in the instance locks table has been updated to be allowed
+	// to start.
+	LockUpdateStartThreshold time.Duration
 }
 
 func (p *PostgresPersister) Delete(ctx context.Context, hash lntypes.Hash) error {
-	result, err := p.conn.ModelContext(ctx, (*dbInvoice)(nil)).
-		Where("hash = ?", hash).Delete() // nolint:contextcheck
+	err := p.conn.RunInTransaction(ctx, func(tx *pg.Tx) error {
+		err := p.checkLockValidity(ctx, tx)
```
joostjager, line 105: I don't think this is sufficient. Multiplexer also interacts with lnd. If we interact with lnd and we've lost the lock without knowing (because we didn't yet do a db tx), there could still be multiple instances.

I believe the general approach is to run the check in an independent loop.

persistence/pg.go
```diff
@@ -241,17 +320,102 @@ func (p *PostgresPersister) Close() error {
...
+	}
+
+	var lock dbInstanceLock
+
+	err := tx.ModelContext(ctx, &lock).Select()
+	if err != nil {
+		return err
+	}
+
+	if lock.LockUpdatedAt != p.lastUpdate {
```
joostjager, line 337: Not sure if this is completely safe. I suppose that in theory another instance could have updated with the exact same time stamp? I don't think you can avoid adding an id to the table as well.

joostjager: I've also added a link to the issue that may contain useful ideas too.

Torakushi: This is not safe indeed, one way could be, to have the UUID or the ID of the mux instance in your table.

Then you could do a:
`SELECT. xxxxx  FOR UPDATE SKIP LOCKED` in a TX when you launch a lnmux instance.
This will prevent any other instance to run (if UUID is not nil or if a Tx has selected the row)

Torakushi: In the rare case of a crash that prevents the instance from removing the lock, we could imagine having a field like your `last_update` in the DB, and the instance which has the lock will update (for update skip lock) this field each minute for example.
If other instances (that are waiting for the lock) see that `last_update` is older than a threshold, then one other instance could acquire the lock.

Torakushi: Imagine the case where an instance didn't update `last_update` but is still alive and another instance acquires the lock. (REALLY UNLIKELY, we don't know why),
You have to check each time you call the database if the UUID is still our UUID. + a context cancelation that will stop the actual worker/server and make us retry to acquire the lock before continuing.

These are only ideas, I might be wrong

persistence/pg.go
Torakushi: If we want to use the PG, we may use Notify feature of PG to know when we release the lock

aakselrod-nydig:
Closing this as we'll be using k8s leases instead.

PR #70, opened 2022-09-05
fix documentation and add a docker-compose

### LINKED ISSUE #40: Document running lnmux in docker
PR #38 added a docker build. What might be useful is a description of how to run lnmux in docker via the command line.

Torakushi:
close #40


README.md
```diff
@@ -110,15 +127,67 @@ If you've set up `lnmuxd` correctly, output similar to what is shown below is ex
...
+
+```bash
+docker build .
+```
+
+**However**, we don't provide any configuration file in the Dockerfile (and don't do the migration either).
+You have various options:
+
+1) Use a `docker-compose.yml` to mount a volume to provide your configuration file, and then override the command instructions.
+   You can take example from the docker-compose we provide for testing purposes. (Explained [bellow](#regtest-testing))
```
joostjager: bellow -> below

README.md
```diff
@@ -110,15 +127,67 @@ If you've set up `lnmuxd` correctly, output similar to what is shown below is ex
...
+2) Copy your configuration using `docker cp` and then connect to the container to launch migrations and run lnmux using:
```
joostjager: I am not sure if it is worth suggesting this. Mounting a config file seems to be the option that is generally used with pre-built docker images.

README.md
```diff
@@ -110,15 +127,67 @@ If you've set up `lnmuxd` correctly, output similar to what is shown below is ex
...
 ## Regtest testing
 
-The minimal setup to test on regtest is to create three LND nodes A, B and C. Create channels between B and A and between B and C. Connect `lnmuxd` to the nodes A and C. Pay invoices from node B, while experimenting with online status and liquidity of A and C.
+The minimal setup to test on regtest is to create three LND nodes Alice, Bob and Carol.
+Alice and Bob should be connected to Carol, and `lnmuxd` connected to Alice and Bob.
+
+The typical workflow is:
+- Create invoices using lnmux
+- Carol pays the invoices
+- Monitor online status and liquidity of Alice and Bob.
```
joostjager: This seems more like a general activity for a lightning setup than that it is specific to lnmux? I'd say some like "lnmux generates accepted events and settles the invoices either through alice or bob"

docker-compose.yml
```diff
@@ -0,0 +1,38 @@
...
+
+services:
+  lnmux:
+    build:
+      context: .
+      target: debian
+    ports:
+      - "9190:9190"
+    volumes:
+      # Copy the configuration file
```
joostjager: I'd call this 'mount', because it isn't copied.

README.md
```diff
@@ -30,62 +30,79 @@ For the setup, it is assumed that there are multiple LND nodes running with conn
...
-  (replace base64 hash)
+  Each time an invoice is accepted, you will receive a pair (hash , setID):
+  
+  ```json
+  Response contents:
+  {
+    "hash": "rCdcdDvpo7oqkO1Gkh1hikofseQpUZhgAIaw6gYXz7M=",
+    "setId": "J1q7YDvixChKrJjjP/yvv16KiQdpHqwW7gy8jqxDAvI="
+  }
+  ```
```
joostjager, line 96: NIce updates here. Would do them in a separate commit, independent of the docker additions.

README.md
```diff
@@ -110,15 +127,67 @@ If you've set up `lnmuxd` correctly, output similar to what is shown below is ex
...
+To build the lnmux images, navigate to the lnmux repository and simply do:
+
+```bash
+docker build .
+```
+
+**However**, we don't provide any configuration file in the Dockerfile (and don't do the migration either).
+You have various options:
+
+1) Use a `docker-compose.yml` to mount a volume to provide your configuration file, and then override the command instructions.
```
joostjager: The original issue didn't mention docker-compose, just a command line how to launch lnmux in docker. I am not sure about the docker-compose addition. It still requires lnd nodes that run on the host, where I'd expect docker-compose to be a completely self-contained setup including lnd nodes and bitcoind. The way you defined it currently, it's just lnmux + postgres. And a path that needs to be changed.

I think it is better to limit ourselves to a `docker run` command line that mounts a config file into the pre-built image.

README.md
Torakushi: Will remove it anyway, since we don't need docker-compose

README.md
```diff
@@ -127,15 +127,65 @@ If you've set up `lnmuxd` correctly, output similar to what is shown below is ex
...
+  ghcr.io/bottlepay/lnmux:sha-bf41263-alpine /bin/sh -c \
+  "/usr/bin/lnmuxd -c /app/lnmux.yml migrate init && \
+   /usr/bin/lnmuxd -c /app/lnmux.yml migrate up && \
+   /usr/bin/lnmuxd -c /app/lnmux.yml run"
+
+```
+
+
+> ⚠️ Don't forget to add **before starting nodes connected to lnmux** (aka Alice and Bob):
+> - `--requireinterceptor` to intercept htlc
```
joostjager: This option is not to enable htlc interception, but to make sure that all htlcs are intercepted.

README.md
```diff
@@ -116,9 +133,59 @@ Note that only the states `accepted`, `settle requested` and `settled` are publi
...
+> - `--tlsextradomain=host.docker.internal` to add docker DSN to TLS certs (otherwise you can't connect from a docker container)
+>
+> If your nodes are already started, you will have to stop them, remove manually tls certs and then restart them.
+> 
+> Do not forget to use `host.docker.internal` instead of `localhost` in your configuration
```
joostjager: The paragraph already starts with 'don't forget', so may not need repeating because the reader is already in not forget mode?

README.md
```diff
@@ -116,9 +133,59 @@ Note that only the states `accepted`, `settle requested` and `settled` are publi
...
+
+**However**, we don't provide any configuration file in the Dockerfile (and don't do the migration either).
+You need to provide certificates as well to connect lnd nodes.
+
+To do so, you can simply use `--mount`:
+
+```bash
+ docker run -it \
+  --mount type=bind,source="$(pwd)/lnmux.yml",target=/app/lnmux.yml \
+  --mount type=bind,source="$(pwd)/pierre.beugnet/lnd/certs",target=/app/certs  \
```
joostjager: You sure you want your real name here?

README.md
```diff
@@ -116,9 +133,59 @@ Note that only the states `accepted`, `settle requested` and `settled` are publi
...
+>
+> If your nodes are already started, you will have to stop them, remove manually tls certs and then restart them.
+> 
+> Do not forget to use `host.docker.internal` instead of `localhost` in your configuration
+
 ## Regtest testing
 
-The minimal setup to test on regtest is to create three LND nodes A, B and C. Create channels between B and A and between B and C. Connect `lnmuxd` to the nodes A and C. Pay invoices from node B, while experimenting with online status and liquidity of A and C.
+The minimal setup to test on regtest is to create three LND nodes Alice, Bob and Carol.
+Alice and Bob should be connected to Carol, and `lnmuxd` connected to Alice and Bob.
```
joostjager: A and B should not just be connected, but also have channels with C.

Torakushi: This is what I meant by `connected` but it is unclear indeed!

README.md
```diff
@@ -127,15 +127,66 @@ If you've set up `lnmuxd` correctly, output similar to what is shown below is ex
 
 ## Invoice lifecycle
 
-![](images/invoice_lifecycle.png)
+![](invoice_lifecycle.png)
```
joostjager: Strange change in this commit?

README.md
```diff
@@ -86,7 +86,7 @@ For the setup, it is assumed that there are multiple LND nodes running with conn
   ````
 
   Each time an invoice is accepted, you will receive a pair (hash , setID):
-
+
```
joostjager: Unrelated to this commit?

joostjager (approved):

README.md
Torakushi: That's a vestige, will fix that

Torakushi: Fix

README.md
Torakushi: Fix white space

PR #77, opened 2022-09-08
mux: remove unused persister field in Mux struct

aakselrod-nydig:
The `persister` field in the `Mux` struct is unused as only the invoice registry actually uses a persister. We may be able to remove it from the `Mux` struct if there are no plans for its future use.


joostjager (approved):

Torakushi (approved):
LGTM

PR #78, opened 2022-09-09
log wait for settled path

joostjager:
If any race condition would appear in the future in this area, we have an additional log line that may help.


Torakushi (approved):
Looks nice

PR #79, opened 2022-09-09
add lnmux_invoice_accepted_subscribers gauge

joostjager:
No application being subscribed to payments that are accepted is usually a bad sign. This PR adds a metric to allow alerting on this via prometheus.


Torakushi (approved):
Tested, works well

PR #80, opened 2022-09-09
add grpc logging

joostjager:


Torakushi (approved):
Looks good.

Talking about interceptor, I add a comment here:
https://github.com/bottlepay/lnmux/issues/15#issuecomment-1241571755

If we really want to monitor our grpc-server

PR #83, opened 2022-09-09
make logger configurable

### LINKED ISSUE #81: Make lnmux logging suitable for prod environment
Currently in lnmux we use the default zap development configuration:

`config := zap.NewDevelopmentConfig()`

This is not suitable for a prod environment and we need to make it configurable (for example, log encoding must be in `json` for a prod environment otherwise severity will not be displayed correctly)

We could for example add in [this](https://github.com/bottlepay/lnmux/pull/80)  `LoggingConfig`  some fields like `Level`, `Format` (console or json), `withEncodeCaller`,...

Torakushi:
close #81

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ For the setup, it is assumed that there are multiple LND nodes running with conn
   * LND nodes configuration: TLS certificate, macaroon, address and pubkey. Pubkey is configured as a protection against unintentionally connecting to the wrong node.
   * Postgres connection string for the database created above (can be overridden by the environment variable `LNMUX_PERSISTENCE_POSTGRES_DSN` to avoid exposing credentials).
   * 32-byte Identity PRIVATE key. This key is used to generate invoices and decode incoming htlcs. In a production environment, this key must be protected carefully (can be overridden by the environment variable `LNMUX_IDENTITY_KEY` to avoid exposing the private key).
+  * Your logging configuration. The most important options are the logs level, the encoding format and the activation of grpc logs.
 
 * Initialize the database: `go run ./cmd/lnmuxd -c lnmux.yml migrate init`
 
--- a/cmd/lnmuxd/config.go
+++ b/cmd/lnmuxd/config.go
@@ -39,13 +39,24 @@ type Config struct {
 	// and eventually pprof and/or other instrumentation listen on.
 	InstrumentationAddress string `yaml:"instrumentationAddress"`
 
-	Logging LoggingConfig `mapstructure:"logging" json:"logging" yaml:"logging"`
+	Logging LoggingConfig `yaml:"logging"`
 
 	DistributedLock DistributedLockConfig `mapstructure:"distributedLock" json:"distributedLock" yaml:"distributedLock"`
 }
 
 // LoggingConfig contains options related to log outputs.
 type LoggingConfig struct {
+	// Level defined the minimum log level to be output
+	Level string `yaml:"level"`
+
+	// Format defined the output format:
+	//  'console' will be human-readable
+	//  'json' will be machine-readable.
+	Format string `yaml:"format"`
+
+	// WithCaller indicates if the caller is displayed in the log
+	WithCaller bool `yaml:"withCaller"`
+
 	// Enables grpc-level logging output.
 	GrpcLogging bool `yaml:"grpcLogging"`
 
--- a/cmd/lnmuxd/log.go
+++ b/cmd/lnmuxd/log.go
@@ -1,16 +1,72 @@
 package main
 
 import (
+	"fmt"
+	"strings"
+
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
 )
 
 var log *zap.SugaredLogger
 
-func init() {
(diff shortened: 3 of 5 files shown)


cmd/lnmuxd/config.go
```diff
@@ -39,11 +39,22 @@ type Config struct {
...
+	// Level defined the minimum log level to be output
+	Level string `yaml:"level"`
+
+	// Format defined the output format:
+	//  'console' will be human-readable
+	//  'json' will be machine-readable.
+	Format string `yaml:"format"`
+
+	// WithCaller indicates if the caller is displayed in the log
+	WithCaller bool `yaml:"withCaller"`
```
Torakushi, line 58: I used this field to keep the actual behaviour as it is.

But if it is not a problem to display the caller in the console, then I will remove it

lnmux.yml.example
```diff
@@ -18,3 +18,9 @@ db:
   # maxConnAge: "15s"
   # poolTimeout: "15s"
   # idleTimeout: "15s"
+logging:
+  # level: "debug"
+  # format: "console"
+  # withCaller: true
+  # grpcLogging: true
+  # grpcLogging: true
```
aakselrod-nydig: This is duplicated

cmd/lnmuxd/config.go
aakselrod-nydig: I like having the option.

aakselrod-nydig (commented):
Looks good, just a duplicated line in the example config.

lnmux.yml.example
```diff
@@ -18,3 +18,9 @@ db:
   # maxConnAge: "15s"
   # poolTimeout: "15s"
   # idleTimeout: "15s"
+logging:
+  # level: "debug"
+  # format: "console"
+  # withCaller: true
+  # grpcLogging: true
+  # grpcPayloadLogging: true
```
aakselrod-nydig, line 26: ⚡

cmd/lnmuxd/log.go
```diff
@@ -1,16 +1,74 @@
...
+	encoderConfig.TimeKey = "time"
+
+	if !withCaller {
+		encoderConfig.EncodeCaller = nil
+	}
+
+	// format is 'console' by default
+	if format == "" {
+		// Add color to the level
+		encoderConfig.EncodeLevel = zapcore.CapitalColorLevelEncoder
```
aakselrod-nydig, line 47: ⚡

aakselrod-nydig (approved):
LGTM

README.md
```diff
@@ -21,6 +21,7 @@ For the setup, it is assumed that there are multiple LND nodes running with conn
   * LND nodes configuration: TLS certificate, macaroon, address and pubkey. Pubkey is configured as a protection against unintentionally connecting to the wrong node.
   * Postgres connection string for the database created above (can be overridden by the environment variable `LNMUX_PERSISTENCE_POSTGRES_DSN` to avoid exposing credentials).
   * 32-byte Identity PRIVATE key. This key is used to generate invoices and decode incoming htlcs. In a production environment, this key must be protected carefully (can be overridden by the environment variable `LNMUX_IDENTITY_KEY` to avoid exposing the private key).
+  * Your logging configuration. You can choose the level, the encoding format, to display grpc logs or not,...
```
joostjager: "or not,..." ?

cmd/lnmuxd/log.go
```diff
@@ -1,16 +1,74 @@
...
+		OutputPaths:      []string{"stderr"},
+		ErrorOutputPaths: []string{"stderr"},
+	}
+
+	lo, err := config.Build()
+	if err != nil {
+		return err
+	}
+
+	lo = lo.With(zap.String("svc", "lnmux"))
```
joostjager: Isn't this unnecessary as there are no other services in this process?

joostjager (commented):
Nice addition

README.md
Torakushi: Maybe it is unclear, I will remove the 'or not'

cmd/lnmuxd/log.go
Torakushi: It is interesting to have this, so we can filter logs given the service name (in an environment where we have various services).

joostjager: Isn't a field for the service name automatically added in such environments by the environment itself? I mean, in k8s for example, not all logs line of all processes are piled on a single heap with any additional identification of where it originates from?

README.md
joostjager: Ah, you mean "etcetera"? Maybe more informative to write it as something like "the most important options are x and y".

Torakushi: Yeah, I mean etcetera, but you are right, your formulation looks better!

cmd/lnmuxd/log.go
Torakushi: Yeah, I guess we can filter per container name in k8s. But maybe we will have some other environment where the service name will be missing? 
I don't really know about other env.
 K8s remains the norm so I can remove it I guess!

joostjager (approved):

PR #86, opened 2022-09-15
remove logs in  for settled or cancelled invoice

### LINKED ISSUE #85: Remove logs in `failInvoice` for settled or cancelled invoice
Currently, even if an invoice is settled, we will still display the log:

`logger.Debugw("Invoice to fail no longer open/accepted")` 

because the invoice is still in `autoReleaseHeap`.

It could be good to remove it from `autoReleaseHeap` or to change the log.

Torakushi:
close #85

### DIFF
--- a/invoiceregistry.go
+++ b/invoiceregistry.go
@@ -489,8 +489,8 @@ func (i *InvoiceRegistry) failInvoice(hash lntypes.Hash) error {
 	// Retrieve invoice.
 	set, ok := i.sets.get(hash)
 	if !ok {
-		logger.Debugw("Invoice to fail no longer open/accepted")
-
+		// Invoice has already been settled or cancelled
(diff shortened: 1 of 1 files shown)


aakselrod-nydig (approved):
LGTM!

invoiceregistry.go
```diff
@@ -489,8 +489,7 @@ func (i *InvoiceRegistry) failInvoice(hash lntypes.Hash) error {
 	// Retrieve invoice.
 	set, ok := i.sets.get(hash)
 	if !ok {
-		logger.Debugw("Invoice to fail no longer open/accepted")
-
+		// Invoice has already been settled or cancelled
```
joostjager, line 492: Could add in comment that we don't log because this is lazy clean up.

joostjager (approved):

PR #88, opened 2022-09-15
start instrumentation server before trying to acquire the k8s lease

### LINKED ISSUE #87: Start instrumentation server before trying to acquire the k8s lease
Currently, we don't start instrumentation server for instances waiting for the lease.

If we want to use some k8s probs, it could be a problem.

Furthermore, we should be able to request statistic for all lnmux instances (even those who wait)

Torakushi:
close #87

### DIFF
--- a/cmd/lnmuxd/run.go
+++ b/cmd/lnmuxd/run.go
@@ -194,41 +194,8 @@ func runAction(c *cli.Context) error {
 		return err
 	}
 
-	instAddress := cfg.InstrumentationAddress
-	if instAddress == "" {
-		instAddress = DefaultInstrumentationAddress
-	}
-
-	// Instantiate a new HTTP server and mux.
-	instMux := http.NewServeMux()
-
-	// Register the Prometheus handler.
-	instMux.Handle("/metrics", promhttp.Handler())
-
-	// Register the pprof handlers. We do this manually because we aren't
-	// using the default mux.
-	// See issues https://github.com/golang/go/issues/42834 and
-	// https://github.com/golang/go/issues/22085
-	instMux.HandleFunc("/debug/pprof", pprof.Index)
-	instMux.HandleFunc("/debug/pprof/cmdline", pprof.Cmdline)
-	instMux.HandleFunc("/debug/pprof/profile", pprof.Profile)
-	instMux.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
-	instMux.HandleFunc("/debug/pprof/trace", pprof.Trace)
-	instMux.Handle("/debug/pprof/allocs", pprof.Handler("allocs"))
-	instMux.Handle("/debug/pprof/block", pprof.Handler("block"))
-	instMux.Handle("/debug/pprof/goroutine", pprof.Handler("goroutine"))
-	instMux.Handle("/debug/pprof/heap", pprof.Handler("heap"))
-	instMux.Handle("/debug/pprof/mutex", pprof.Handler("mutex"))
-	instMux.Handle("/debug/pprof/threadcreate", pprof.Handler("threadcreate"))
-
-	instServer := &http.Server{
-		Addr:    instAddress,
-		Handler: instMux,
-
-		// Even though this server should only be exposed to trusted
-		// clients, this mitigates slowloris-like DoS attacks.
-		ReadHeaderTimeout: 10 * time.Second,
-	}
+	// Initialise instrumentation server
+	instServer := initInstrumentationServer(cfg.InstrumentationAddress)
 
 	group, ctx := errgroup.WithContext(context.Background())
 
@@ -265,7 +232,7 @@ func runAction(c *cli.Context) error {
 
 	group.Go(func() error {
 		log.Infow("Instrumentation HTTP server starting",
-			"instrumentationAddress", instAddress)
+			"instrumentationAddress", instServer.Addr)
 
 		return instServer.ListenAndServe()
 	})
@@ -396,4 +363,41 @@ func initDistributedLock(cfg *DistributedLockConfig) (func(), error) {
 		DevKubeConfig: cfg.DevKubeConfig,
 		Logger:        log,
 	})
+}
+
+func initInstrumentationServer(instAddress string) *http.Server {
+	if instAddress == "" {
+		instAddress = DefaultInstrumentationAddress
+	}
+
+	// Instantiate a new HTTP server and mux.
+	instMux := http.NewServeMux()
+
+	// Register the Prometheus handler.
+	instMux.Handle("/metrics", promhttp.Handler())
+
+	// Register the pprof handlers. We do this manually because we aren't
+	// using the default mux.
+	// See issues https://github.com/golang/go/issues/42834 and
+	// https://github.com/golang/go/issues/22085
+	instMux.HandleFunc("/debug/pprof", pprof.Index)
+	instMux.HandleFunc("/debug/pprof/cmdline", pprof.Cmdline)
+	instMux.HandleFunc("/debug/pprof/profile", pprof.Profile)
+	instMux.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
+	instMux.HandleFunc("/debug/pprof/trace", pprof.Trace)
+	instMux.Handle("/debug/pprof/allocs", pprof.Handler("allocs"))
+	instMux.Handle("/debug/pprof/block", pprof.Handler("block"))
+	instMux.Handle("/debug/pprof/goroutine", pprof.Handler("goroutine"))
+	instMux.Handle("/debug/pprof/heap", pprof.Handler("heap"))
+	instMux.Handle("/debug/pprof/mutex", pprof.Handler("mutex"))
+	instMux.Handle("/debug/pprof/threadcreate", pprof.Handler("threadcreate"))
+
+	return &http.Server{
+		Addr:    instAddress,
+		Handler: instMux,
+
+		// Even though this server should only be exposed to trusted
+		// clients, this mitigates slowloris-like DoS attacks.
+		ReadHeaderTimeout: 10 * time.Second,
+	}
 }
--- a/cmd/lnmuxd/run.go
+++ b/cmd/lnmuxd/run.go
@@ -48,6 +48,26 @@ func runAction(c *cli.Context) error {
 		return err
 	}
 
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	group, ctx := errgroup.WithContext(ctx)
+
+	err = run(ctx, cfg, group)
+	if err != nil {
+		cancel()
+
+		// We don't need to catch the error since we only want to wait all goroutines to finish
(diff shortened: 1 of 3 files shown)


cmd/lnmuxd/run.go
```diff
@@ -48,6 +49,27 @@ func runAction(c *cli.Context) error {
 		return err
 	}
 
+	group, ctx := errgroup.WithContext(context.Background())
+
+	// Initialise instrumentation server
+	instServer := initInstrumentationServer(cfg.InstrumentationAddress)
+	group.Go(func() error {
```
Torakushi, line 89: Is there a reason to use errGroup instead if waitGroup?

I stick with errGroup, but it is less flexible here.

Here we have a problem if we shutdown the process for an instance that waits for the lease.
Indeed, if we shutdown the process during `initDistributedLock`, we will not shutdown gracefully the metrics server.

2 solutions:

- use waitGroup for this one (but it is not really consistent w/ errGroup)
- we don't care, closing gracefully the instrumentation server is not really critical. Furthermore, this scenario is not usual.


This is not the normal workflow, and the same thing could happen if we interrupt the process before reaching errGroup.Wait(). 
In all these **_unlikely_** scenarios, the scenario in which we stop the process when trying to acquire the lock is more **_likely_** :)

aakselrod-nydig: I lean toward solution 2, we don't really care if the instrumentation server closes gracefully.

cmd/lnmuxd/run.go
```diff
@@ -48,6 +48,27 @@ func runAction(c *cli.Context) error {
...
+		// Stop instrumentation server
+		log.Infow("Instrumentation server stopping")
+
+		return instServer.Close()
+	})
+
+	// Setup kubernetes lock
 	releaseLock, err := initDistributedLock(&cfg.DistributedLock)
 	if err != nil {
 		return err
```
aakselrod-nydig, line 108: Another option to enforce clean shutdown in previously-started goroutines is, before this return, to do the following:

```
group.Go(func() error {return err})
```

cmd/lnmuxd/run.go
```diff
@@ -48,6 +48,27 @@ func runAction(c *cli.Context) error {
...
 	}
 
+	group, ctx := errgroup.WithContext(context.Background())
```
aakselrod-nydig: Another option (I prefer this one) would be to change this to something like:
```
backCtx, cancel := context.WithCancel(context.Background())
defer cancel()

group, ctx := errgroup.WithContext(backCtx)
```
and then if we return early, we cancel the error group.

aakselrod-nydig (approved):

cmd/lnmuxd/run.go
```diff
@@ -397,3 +364,40 @@ func initDistributedLock(cfg *DistributedLockConfig) (func(), error) {
 		Logger:        log,
 	})
 }
+
+func initInstrumentationServer(instAddress string) *http.Server {
```
joostjager, line 386: For review, it would be useful to have this method extraction in a separate preparatory commit.

cmd/lnmuxd/run.go
joostjager: We used waitgroup before, but it can get complicated with waiting for all the goroutine results and keeping track of what to terminate.

If `initDistrbutedLock` returns an error, couldn't you cancel the context and `group.Wait()`?

cmd/lnmuxd/run.go
Torakushi: The problem remains the same no? 
Since we don't have a `errgroup.Wait()` we will not wait for all errgroup goroutines (so we will not wait the server to close gracefully)

cmd/lnmuxd/run.go
joostjager: Oh, I see this is what @aakselrod-nydig suggests, but then with an added wait

cmd/lnmuxd/run.go
Torakushi: I see what you mean, but the real problem is that the errgroup.Wait() is called only at the end, right?

cmd/lnmuxd/run.go
Torakushi: Yeah we could do that,

I found this strange to do this only here. 
Why not doing this for `mux,err := lnmux.New(&lnmux.MuxConfig{...})`, etc... 

But I suppose, as I said, this is the most likely of unlikely scenarios! So it could make sense!

joostjager: I see. How about `defer group.Wait()` ?

Torakushi: Was thinking of this indeed, but we want to return the value of `group.Wait()` (if any error)
And we can't do this in a defer.

But we can have two group.Wait(), one defer and the other one at the end at the function. A bit strange but it is okay for me!

(as a note: group.Wait() is not idempotent)

joostjager: You can return that `group.Wait` result if you make it a named return parameter I think.

Torakushi: Indeed, it is less readable but It avoids to use 2 group.Wait()

joostjager: Or indeed just do something special only for the admin server and don't include it in the global errgroup.

cmd/lnmuxd/run.go
```diff
@@ -48,6 +48,50 @@ func runAction(c *cli.Context) error {
 		return err
 	}
 
+	// Initialise instrumentation server
+	instServer := initInstrumentationServer(cfg.InstrumentationAddress)
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
```
joostjager, line 52: Is this still necessary with the defer below? Either the group finished, or we already called `cancel`.

cmd/lnmuxd/run.go
```diff
@@ -48,6 +48,50 @@ func runAction(c *cli.Context) error {
...
+
+	group, ctx := errgroup.WithContext(ctx)
+	// Wait for all goroutine to finish.
+	// Return 'group.Wait()' error if no previous error was encountered.
+	defer func() {
```
joostjager: Looking at this now, I wonder if things can be made more readable by having a `runAction` that calls into an inner `run(group) error`? To avoid the named parameter and sophisticated defer? Maybe it is too much though, not sure.

cmd/lnmuxd/run.go
```diff
@@ -157,7 +157,7 @@ func run(ctx context.Context, cfg *Config, group *errgroup.Group) error {
 			grpc_zap.UnaryServerInterceptor(log.Desugar()),
 		)
 		streamInterceptors = append(streamInterceptors,
-			grpc_zap.StreamServerInterceptor(log.Desugar()),
+			grpc_zap.StreamServerInterceptor(log.Desugar()), //nolint: contextcheck
```
Torakushi, line 204: Add //nolint: contextcheck otherwise the linter will detect a problem we can't solve... 
https://github.com/bottlepay/lnmux/actions/runs/3090768021/jobs/4999993946

dlock/dlock.go
```diff
@@ -132,20 +133,25 @@ func New(cfg *LockConfig) (func(), error) {
...
-		// the main thread my already terminate the process before the
-		// lock is released.
-		wg.Wait()
+			logger.Infow("Lock released")
+		}
+
+		return unlock, nil
 
-		logger.Infow("Lock released")
+	case <-ctx.Done():
```
joostjager, line 153: Just to be sure: if the external context is cancelled, there is no way that this function can return no error, right?

cmd/lnmuxd/run.go
```diff
@@ -48,6 +48,17 @@ func runAction(c *cli.Context) error {
 		return err
 	}
 
+	group, ctx := errgroup.WithContext(context.Background())
+
+	group.Go(func() error {
```
joostjager: I still don't fully understand why this needs to run in a goroutine. Wouldn't something like this work?

```golang
err := run(ctx, cfg, group)
if err != nil {
   cancel()

  // Error can be ignored because we already have the primary error
   _ = group.Wait()

  return err
}

return group.Wait()
```

Maybe it comes down to the same thing though...

joostjager: Just for this refactor commit, I think the extra goroutine isn't necessary yet? It would be sufficient to just `group.Wait` if err == nil, in line with the original code? Then add the additional handling in the last commit.

Torakushi: Yes indeed it works this way, it was my first approach.

But using the goroutine allows to avoid the double `group.Wait()` (I know it is in different paths but.. still).

But if you think this is more readable, I will go with this approach!

dlock/dlock.go
Torakushi: No, if the context is cancelled we just stop `RunOrDie`, then we will block on <-lockacquired (that's why I use now a `select` statement)

cmd/lnmuxd/run.go
joostjager: It would be my preference to avoid the nested goroutine, for readability

Torakushi: Done! I added this modif in the last commit

cmd/lnmuxd/run.go
```diff
@@ -48,20 +48,75 @@ func runAction(c *cli.Context) error {
...
+	defer cancel()
+
+	group, ctx := errgroup.WithContext(ctx)
+
+	err = run(ctx, cfg, group)
+	if err != nil {
+		cancel()
+
+		// We don't need to catch the error since we only want to wait all goroutines to finish
+		group.Wait() // nolint: errcheck
```
joostjager: If you do `_ = group.Wait()`, doesn't the linter error go away?

Torakushi: Yes, indeed.

Usually I use // nolint: errcheck instead of `_ = xxxxx` (since we want to avoid linter error), but seeing your comment you prefer the other way!
I will change it

joostjager: Yes, I think generally that is the preferred way. To solve it in code.

Torakushi: done

cmd/lnmuxd/run.go
joostjager: I suppose this is indeed necessary just to keep the linter quiet?

dlock/dlock.go
```diff
@@ -132,20 +133,25 @@ func New(cfg *LockConfig) (func(), error) {
...
-	logger.Debugw("Lock acquired")
+		unlock := func() {
+			cancel()
 
-	unlock := func() {
-		cancel()
+			// Wait for `leaderelection.RunOrDie` to terminate. Otherwise
+			// the main thread my already terminate the process before the
+			// lock is released.
+			wg.Wait()
```
joostjager, line 146: This one is not called anymore now in the `ctx.Done()` branch?

joostjager (approved):
A lot of extra discoveries in this PR, but I think worth it!

PR #91, opened 2022-09-23
Fix dlock behaviour and error handling 

Torakushi:
2 fixes:

First, we need to do `errgroup.Wait()` in `run` function otherwise defer statements will be triggered. For example, if an instance acquires the lock, it will reach the end of `run` function and then trigger the `defer releaseLock()` (and another instance can acquire the lock)

Secondly, in this code:

```
func run(){

group.Go(func() error {
		log.Infow("Instrumentation HTTP server starting",
			"instrumentationAddress", instServer.Addr)
		return instServer.ListenAndServe()
	})
	
releaseLock, err := initDistributedLock(ctx, &cfg.DistributedLock)
	if err != nil {
		return err
	}
	
return group.Wait()

}
```

If we have an error when calling `instServer.ListenAndServe()`, we will return the error `context canceled` instead of the ListenAndServe error.
Indeed, the `ctx` in `initDistributedLock` is the `errgroup` context.
So , we will cancel the context if an error occurs in a Group.Go(), and then, initDistributedLock will return `context canceled`.

### DIFF
--- a/cmd/lnmuxd/run.go
+++ b/cmd/lnmuxd/run.go
@@ -48,26 +48,14 @@ func runAction(c *cli.Context) error {
 		return err
 	}
 
-	ctx, cancel := context.WithCancel(context.Background())
-	defer cancel()
-
-	group, ctx := errgroup.WithContext(ctx)
-
-	err = run(ctx, cfg, group)
-	if err != nil {
-		cancel()
-
-		// We don't need to catch the error since we only want to wait all goroutines to finish
-		_ = group.Wait()
-
-		return err
-	}
-
-	return group.Wait()
+	return initServiceWithLock(cfg.InstrumentationAddress, cfg.DistributedLock, func(ctx context.Context) error {
+		return run(ctx, cfg)
+	})
 }
 
-func run(ctx context.Context, cfg *Config, group *errgroup.Group) error {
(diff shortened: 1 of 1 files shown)


Torakushi:
I agree... this is less readable (I added comment to make it clearer)

Torakushi:
Another solution could be:
 
Put the `run` function into a group.Go and make `run` return the `releaseLock()` function so we can do

```
group.Go(func() error {
	  releaseLock, err :=  run(ctx, cfg, group)
          if releaseLock != nil{
                 releaseLock()
           }
	 
	  return err
	})
```


EDIT: 

I think I prefer this solution finally... What do you think?

cmd/lnmuxd/run.go
```diff
@@ -53,8 +53,10 @@ func runAction(c *cli.Context) error {
 
 	group, ctx := errgroup.WithContext(ctx)
 
+	// If the error is not "context canceled", then we either reached the end of 'run' function
+	// or we got an error which was not thrown by errgroup.
 	err = run(ctx, cfg, group)
-	if err != nil {
+	if err != context.Canceled {
```
joostjager: if `err == nil` we do another `group.Wait`?

joostjager (approved):

PR #95, opened 2022-10-06
grace period during which lnmux waits for a new subscriber

### LINKED ISSUE #94: Do not send ResultNoAcceptSubscriber if the last subscriber disconnected recently
Currently we will send a FailResolution with 'ResultNoAcceptSubscriber' if we try to accept a new set without any subscriber.

It might be good to have a period where lnmux is still accepting new set_ids because it is waiting for a subscriber to reconnect.

after this period, we will send ResultNoAcceptSubscriber for each new set_id

Torakushi:
close #94

### DIFF
--- a/invoiceregistry.go
+++ b/invoiceregistry.go
@@ -73,6 +73,10 @@ type RegistryConfig struct {
 	PrivKey [32]byte
 
 	AutoSettle bool
+
+	// GracePeriodWithoutSubscribers is the default period during which the subscriberManager is waiting for a subscriber
+	// to reconnect.
+	GracePeriodWithoutSubscribers time.Duration
 }
 
 type InvoiceCallback func(update InvoiceUpdate)
@@ -135,7 +139,7 @@ func NewRegistry(cdb *persistence.PostgresPersister,
 		htlcChan:                        make(chan *registryHtlc),
 		newInvoiceAcceptSubscription:    make(chan invoiceAcceptSubscription),
 		cancelInvoiceAcceptSubscription: make(chan int),
-		subscriptionManager:             newSubscriptionManager(cfg.Logger),
+		subscriptionManager:             newSubscriptionManager(cfg.Logger, cfg.GracePeriodWithoutSubscribers),
 		requestSettleChan:               make(chan *invoiceRequest),
 		cancelChan:                      make(chan *invoiceRequest),
 		quit:                            make(chan struct{}),
@@ -751,12 +755,14 @@ func (i *InvoiceRegistry) process(ctx context.Context, h *registryHtlc) error {
 		// Don't start a new set if there is currently no application listening
 		// for accept events. This is to prevent htlcs being held on to until
 		// the mpp timeout when no one is listening.
+		// If the last disconnection occurred recently, then keep accepting HTLCs
+		// in the hope that a subscriber will reconnect.
 		//
 		// When we are adding to an existing set, we keep going and assume that
 		// the application will be back online soon.
 		//
 		// This restriction does not apply in auto-settle mode.
-		if !i.cfg.AutoSettle && !i.subscriptionManager.hasSubscribers() {
+		if !i.cfg.AutoSettle && !i.subscriptionManager.hasRecentSubscribers() {
 			h.resolve(NewFailResolution(
 				ResultNoAcceptSubscriber,
 			))
--- a/invoiceregistry.go
+++ b/invoiceregistry.go
@@ -170,9 +170,6 @@ func (i *InvoiceRegistry) SubscribeAccept(callback AcceptCallback) (
 	func(), error) {
 
 	subscriberId := i.subscriptionManager.generateSubscriptionId()
-	logger := i.logger.With("id", subscriberId)
-
-	logger.Debugw("Adding subscriber")
 
 	select {
 	case i.newInvoiceAcceptSubscription <- invoiceAcceptSubscription{
@@ -185,8 +182,6 @@ func (i *InvoiceRegistry) SubscribeAccept(callback AcceptCallback) (
 	}
 
 	return func() {
-		logger.Debugw("Removing subscriber")
-
 		select {
 		case i.cancelInvoiceAcceptSubscription <- subscriberId:
 		case <-i.quit:
--- a/invoiceregistry_test.go
+++ b/invoiceregistry_test.go
@@ -64,6 +64,14 @@ func optAcceptTimeout(d time.Duration) regTestContextOption {
 	}
 }
 
+// optGracePeriodWithoutSubscribers returns a functional argument to newRegistryTestContext()
+// that modifies the gracePeriodWithoutSubscribers configuration in the RegistryConfig.
+func optGracePeriodWithoutSubscribers(d time.Duration) regTestContextOption {
+	return func(c *registryTestContext) {
+		c.cfg.GracePeriodWithoutSubscribers = d
+	}
+}
+
 // optTestAmt returns a functional argument to newRegistryTestContext()
 // that modifies the testAmt configuration.
 func optTestAmt(amt int64) regTestContextOption { // nolint:unused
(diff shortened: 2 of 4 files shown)


invoiceregistry.go
```diff
@@ -751,12 +751,16 @@ func (i *InvoiceRegistry) process(ctx context.Context, h *registryHtlc) error {
...
+		// in the hope that a subscriber will reconnect.
 		//
 		// When we are adding to an existing set, we keep going and assume that
 		// the application will be back online soon.
 		//
 		// This restriction does not apply in auto-settle mode.
-		if !i.cfg.AutoSettle && !i.subscriptionManager.hasSubscribers() {
+		if !i.cfg.AutoSettle &&
+			!i.subscriptionManager.hasSubscribers() &&
+			!i.subscriptionManager.isWaitingForReconnection() {
```
joostjager: Newline after multi-line stmt

subscription_manager.go
```diff
@@ -38,6 +44,10 @@ func (s *subscriptionManager) deleteAcceptSubscription(id int) {
 	}
 
 	delete(s.acceptSubscriptions, id)
+
+	if len(s.acceptSubscriptions) == 0 {
+		s.lastSubscriberRemovingAt = time.Now()
```
joostjager: Removing -> Removed?

subscription_manager.go
```diff
@@ -53,3 +63,8 @@ func (s *subscriptionManager) notifySubscribers(hash lntypes.Hash, setID [32]byt
 func (s *subscriptionManager) hasSubscribers() bool {
 	return len(s.acceptSubscriptions) > 0
 }
+
+func (s *subscriptionManager) isWaitingForReconnection() bool {
```
joostjager: Maybe this and `hasSubscribers` can be merged into a single function? Perhaps name it something more abstract like `hasRecentSubscribers()`?

joostjager (commented):
Straight-forward change

subscription_manager.go
Torakushi: Yep indeed

subscription_manager.go
Torakushi: yes, it's a vestige

subscription_manager.go
```diff
@@ -50,6 +60,7 @@ func (s *subscriptionManager) notifySubscribers(hash lntypes.Hash, setID [32]byt
 	}
 }
 
-func (s *subscriptionManager) hasSubscribers() bool {
-	return len(s.acceptSubscriptions) > 0
+func (s *subscriptionManager) hasRecentSubscribers() bool {
+	return len(s.acceptSubscriptions) > 0 ||
+		(!s.lastSubscriberRemovedAt.IsZero() && time.Since(s.lastSubscriberRemovedAt) <= gracePeriodWithoutSubscribers)
```
joostjager: Is that `IsZero` check necessary? If timestamp is zero, than the second clause is always false anyway?

Torakushi: it was needed before the merge. I can remove it now, my bad

joostjager (commented):
Not sure if you want to add test coverage at all?

Torakushi:
Good point. I was in a hurry, so I didn't take the time. I will see how I can create tests for this

invoiceregistry_test.go
```diff
@@ -320,6 +320,44 @@ func TestNoSubscriberFail(t *testing.T) {
...
+	c := newRegistryTestContext(t, optAcceptTimeout(0))
+
+	// Add invoice.
+	invoice, preimage := c.createInvoice(1, time.Hour)
+
+	// A subscriber connects and disconnects
+	_, acceptCancel := c.subscribeAccept()
+	acceptCancel()
+
+	require.True(t, c.registry.subscriptionManager.hasRecentSubscribers())
```
Torakushi, line 340: Not necessary, only an extra security

invoiceregistry_test.go
```diff
@@ -320,6 +320,39 @@ func TestNoSubscriberFail(t *testing.T) {
...
 }
 
+// TestGracePeriodForSubscriber asserts that a new set is started even if the last subscriber has recently disconnected.
+// After the grace period, we should not accept any new sets
+func TestGracePeriodForSubscriber(t *testing.T) {
+	defer test.Timeout()()
+	t.Parallel()
+
+	// If we reach the accept timeout, it means we've accepted the set_id
+	c := newRegistryTestContext(t, optAcceptTimeout(0))
```
Torakushi: I used AcceptTimeout instead of directly checking if our setid is in `htlcSets` because I don't like
Time.Sleep() in tests

invoiceregistry_test.go
```diff
@@ -320,6 +320,39 @@ func TestNoSubscriberFail(t *testing.T) {
...
+	_, acceptCancel := c.subscribeAccept()
+	acceptCancel()
+
+	c.notifyHtlc(preimage.Hash(), invoice.PaymentAddr)
+
+	// The set was accepted but we reach accept timeout
+	c.checkHtlcFailed(ResultAcceptTimeout)
+
+	// Grace period is finished
+	gracePeriodWithoutSubscribers = 0
```
joostjager: Is it really necessary to change this value which is mostly a constant?

invoiceregistry_test.go
```diff
@@ -320,6 +320,39 @@ func TestNoSubscriberFail(t *testing.T) {
 	c.checkHtlcFailed(ResultNoAcceptSubscriber)
 }
 
+// TestGracePeriodForSubscriber asserts that a new set is started even if the last subscriber has recently disconnected.
+// After the grace period, we should not accept any new sets
+func TestGracePeriodForSubscriber(t *testing.T) {
```
joostjager: There was already a test for no subscribers, `TestNoSubscriberFail`. I think this may overlap partially with this test?

invoiceregistry_test.go
joostjager: Not completely sure about the test as is. I understand your wish to not use sleep, but it seems you now rely on workarounds that may not be the easiest to follow.

Perhaps it's time to add a fake clock to mux now. It's in the `lnd/clock` package.

invoiceregistry_test.go
Torakushi: It was simply to avoid using time.sleep()

But related to https://github.com/bottlepay/lnmux/pull/95#discussion_r989923982

Using a clock will solve the problem

invoiceregistry_test.go
Torakushi: First I thought, it should be 1 test, 1 feature

TestNoSubscriberFail --> test when no subscribers were connected

TestGracePeriodForSubscriber --> test graceful period

But you are right, we can consider this as the same feature maybe

I will merge it

subscription_manager.go
```diff
@@ -2,24 +2,36 @@ package lnmux
...
 
-	nextSubscriberId uint64
+	nextSubscriberId              uint64
+	lastSubscriberRemovedAt       time.Time
+	gracePeriodWithoutSubscribers time.Duration
 }
 
-func newSubscriptionManager(logger *zap.SugaredLogger) *subscriptionManager {
+func newSubscriptionManager(logger *zap.SugaredLogger, gracePeriodWithoutSubscribers time.Duration) *subscriptionManager {
+	if gracePeriodWithoutSubscribers == 0 {
```
joostjager: I wouldn't do this. Defaults on various levels may get confusing. I'd just instantiate the invoice registry with the default value at a higher level.

joostjager (commented):
Test looks good.

subscription_manager.go
Torakushi: Done

subscription_manager.go
```diff
@@ -38,6 +46,10 @@ func (s *subscriptionManager) deleteAcceptSubscription(id int) {
 	}
 
 	delete(s.acceptSubscriptions, id)
+
+	if len(s.acceptSubscriptions) == 0 {
```
joostjager: This if is probably not even required. The field name still checks out when it is updated with every removal.

joostjager (approved):

joostjager (approved):

PR #97, opened 2022-10-21
allow non-zero routing policy

### LINKED ISSUE #93: Add ability to configure virtual channel fees
Currently the fee for the final virtual hop is always zero. We could expose the fee schedule as a setting so that users can charge an inbound fee for getting paid. The fee could be different per connected node, but not sure if there is a use case for that. Perhaps steering traffic?

joostjager:
Allows a routing policy to be set through configuration. The policy is communicated through hop hints and enforced for
intercepted htlcs.

With this change a fee can be charged for receiving payments.

Fixed #93


Torakushi (approved):
Good catch for the deadlock + cool feature

PR #104, opened 2022-10-28
add RPC call to retrieve invoices

### LINKED ISSUE #92: Add settled invoices query rpc
The settle flow in lnmux is designed such that a connecting application is fully in control. It can keep track of all settled invoices by itself in a safe way.

Lnmux however is also storing all settled invoices so that it can handle potential replays of htlcs that come in later. Currently it is not possible to query that dataset from lnmux and a direct db query is needed.

This issue is about adding an rpc to run queries on the settled invoices. To be decided is what the best way is to do this (think pagination etc).

Torakushi:
close #92


persistence/pg.go
```diff
@@ -120,6 +158,22 @@ func (p *PostgresPersister) Get(ctx context.Context, hash lntypes.Hash) (*Invoic
 	return invoice, htlcs, nil
 }
 
+func (p *PostgresPersister) GetSettledInvoices(ctx context.Context, numMaxInvoices, invoiceOffset int) ([]*dbInvoice, error) {
+	var dbInvoices []*dbInvoice
+	err := p.conn.RunInTransaction(ctx, func(tx *pg.Tx) error {
+		return tx.ModelContext(ctx, &dbInvoices).
+			Where("sequence_num>?", invoiceOffset).
+			Where("settled = true").
+			Limit(numMaxInvoices).
```
joostjager: Order?

persistence/migrations/2_add_invoice_id.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE lnmux.invoices ADD "sequence_num" SERIAL;
```
joostjager: UNIQUE constraint? And need to think about the upgrade path for existing invoices. Maybe just drop everything.

lnmuxrpc/lnmux.proto
```diff
@@ -73,3 +77,23 @@ message CancelInvoiceRequest {
...
 message CancelInvoiceResponse {}
+
+message ListInvoiceRequest {
+    uint64 num_max_invoices = 1;
+    uint64 index_offset = 2;
+}
+
+message ListInvoiceResponse {
+    repeated Invoice invoice = 1;
+    uint64 first_index_offset = 2;
```
joostjager: Isn't this just the request index passed back?

lnmuxrpc/lnmux.proto
```diff
@@ -73,3 +77,23 @@ message CancelInvoiceRequest {
...
+    uint64 last_index_offset = 3;
```
joostjager: This can be derived via `index_offset + len(invoice)?`

I am not sure if we should add these convenience fields or not. An api like this has probably been defined thousands of times in the wider world, but I have no idea what's the common way to do it.

cmd/lnmuxd/server.go
```diff
@@ -23,24 +24,30 @@ const (
 	// subscriberQueueSize is the length of the rpc subscriber queue. If the
 	// queue becomes full, the client is disconnected.
 	subscriberQueueSize = 100
+
+	// defaultPaginationSize is the default number of invoices returned by the pagination
+	defaultPaginationSize = 100
```
joostjager: Maybe we should do no default, to force the caller to think about it.

joostjager (commented):
Pretty straight-fwd change, left some comments

lnmuxrpc/lnmux.proto
Torakushi: From what I see using
```
 uint64 first_index_offset = 2;
 uint64 last_index_offset = 3;
```

Is pretty common in the wider world

Imagine we did some manual intervention in the DB: we canceled/deleted invoices with sequence_number == 1 and 3

If I request:

REQUEST( index_off_set=1, max_invoices=3)

I will get [2,4,5]

index_offset != first_index_offset
and
last_index_offset != index_offset + len(invoices)

lnmuxrpc/lnmux.proto
Torakushi: I tried to answer that here:
https://github.com/bottlepay/lnmux/pull/104#discussion_r1009174857

cmd/lnmuxd/server.go
Torakushi: Interesting, Usually (in the wider world), there is a default, but It  is a good idea to "empower" the user

lnmuxrpc/lnmux.proto
joostjager: I see. Ok, let's keep it indeed.

cmd/lnmuxd/server.go
joostjager: With lnd, we've had quite some instances of users not using the api correctly because of defaults. The most notable default was the payment fee limit at 100% of the payment amount.

lnmuxrpc/lnmux.proto
```diff
@@ -73,3 +77,23 @@ message CancelInvoiceRequest {
 }
 
 message CancelInvoiceResponse {}
+
+message ListInvoiceRequest {
+    uint64 num_max_invoices = 1;
```
joostjager: Nit: is `max_invoices_count` better? Or something else?

I am currently reading `Number of maximum invoices', which doesn't parse perfectly.

lnmuxrpc/lnmux.proto
```diff
@@ -73,3 +77,23 @@ message CancelInvoiceRequest {
...
+}
+
+message Invoice {
+    bytes hash = 1;
+    bytes preimage = 2;
+    uint64 amount_msat = 3;
+    bool settled = 4;
+    uint64 settled_requested_at = 5;
+    uint64 settled_at = 6;
+}
```
joostjager, line 100: Add the sequence number here too?

lnmuxrpc/lnmux.proto
```diff
@@ -73,3 +77,23 @@ message CancelInvoiceRequest {
 }
 
 message CancelInvoiceResponse {}
+
+message ListInvoiceRequest {
+    uint64 num_max_invoices = 1;
+    uint64 index_offset = 2;
```
joostjager: Perhaps choose one name: either index or sequence.

persistence/pg.go
```diff
@@ -37,6 +38,9 @@ type InvoiceCreationData struct {
 type dbInvoice struct {
 	tableName struct{} `pg:"lnmux.invoices,discard_unknown_columns"` // nolint
 
+	// SequenceNum is a unique identifier used for pagination
+	SequenceNum uint64 `pg:"sequence_num,pk"`
```
joostjager: Storing the sequence number and adding the query rpc can be separate commits.

joostjager: Hmm, the primary key is defined differently here (seq+hash) than it is in the sql script.

persistence/pg.go
```diff
@@ -82,6 +86,19 @@ func unmarshallDbInvoice(invoice *dbInvoice) *Invoice {
 	}
 }
 
+func DBInvoiceToProto(invoice *dbInvoice) *lnmuxrpc.Invoice {
```
joostjager: Maybe this can remain private, as we don't want to expose `dbInvoice` anyway.

persistence/pg.go
```diff
@@ -82,6 +86,19 @@ func unmarshallDbInvoice(invoice *dbInvoice) *Invoice {
...
+	preimage := invoice.Preimage.Hash()
+
+	return &lnmuxrpc.Invoice{
+		Hash:               invoice.Hash[:],
+		Preimage:           preimage[:],
+		AmountMsat:         uint64(invoice.AmountMsat),
+		Settled:            invoice.Settled,
+		SettledRequestedAt: uint64(invoice.SettleRequestedAt.Unix()),
```
joostjager: Losing precision here I think, but maybe nbd

persistence/pg.go
```diff
@@ -131,6 +148,23 @@ func (p *PostgresPersister) Get(ctx context.Context, hash lntypes.Hash) (*Invoic
 	return invoice, htlcs, nil
 }
 
+func (p *PostgresPersister) GetSettledInvoices(ctx context.Context, numMaxInvoices, invoiceOffset int) ([]*dbInvoice, error) {
+	var dbInvoices []*dbInvoice
+	err := p.conn.RunInTransaction(ctx, func(tx *pg.Tx) error {
+		return tx.ModelContext(ctx, &dbInvoices).
+			Where("sequence_num>?", invoiceOffset).
```
joostjager: `invoiceOffset` is different again from index and sequence. Maybe `sequenceStart`?

persistence/pg.go
```diff
@@ -131,6 +148,23 @@ func (p *PostgresPersister) Get(ctx context.Context, hash lntypes.Hash) (*Invoic
...
+			Where("settled = true").
```
joostjager: Don't we want to allow querying invoices for which settlement has been requested? This is different from invoice that are only accepted.

Torakushi: Will add invoices for which settlement has been requested

persistence/pg.go
Torakushi: Right, I will remove the pk from Sequence number

persistence/pg_test.go
```diff
@@ -152,3 +152,107 @@ func TestConcurrentHtlcSettle(t *testing.T) {
...
+	dbInvoices, err = persister.GetSettledInvoices(context.Background(), 3, 0)
+	require.NoError(t, err)
+	require.Len(t, dbInvoices, 3)
+	require.Equal(t, dbInvoices[0].SequenceNum, uint64(1))
+	require.Equal(t, dbInvoices[0].Preimage, invoices[0].PaymentPreimage)
+	require.Equal(t, dbInvoices[1].SequenceNum, uint64(2))
+	require.Equal(t, dbInvoices[1].Preimage, invoices[1].PaymentPreimage)
+	require.Equal(t, dbInvoices[2].SequenceNum, uint64(3))
+	require.Equal(t, dbInvoices[2].Preimage, invoices[2].PaymentPreimage)
+}
```
joostjager, line 245: Maybe a nice touch to add an assertion to `mux_test.go` where we query settled invoices. To give it a bit more of an e2e character.

persistence/pg.go
Torakushi: Actually, we use DBInvoiceToProto in server.go
We need to export this function

Otherwise we need to return a []lnmuxrpc.Invoice in `persistence` package but it doesn't make sense IMO?

or I can move this function in server.go, but I thought that it should be in the persistence package because we are working on a `dbInvoice`

joostjager: Can't we just return `[]*pg.Invoice`?

Torakushi: you mean for `func (p *PostgresPersister) GetSettledInvoices()` call?

joostjager: yes

Torakushi: Ah yeah we can, so I can move the proto conversion in server.go

persistence/pg.go
```diff
@@ -37,6 +37,9 @@ type InvoiceCreationData struct {
 type dbInvoice struct {
 	tableName struct{} `pg:"lnmux.invoices,discard_unknown_columns"` // nolint
 
+	// SequenceNum is a unique identifier used for pagination
+	SequenceNum uint64 `pg:"sequence_num"`
```
joostjager, line 43: Not 100% sure, but maybe `use_zero` is a safe thing to add here. Annoying that default conversion that go-pg does.

persistence/pg.go
```diff
@@ -37,6 +37,9 @@ type InvoiceCreationData struct {
 type dbInvoice struct {
 	tableName struct{} `pg:"lnmux.invoices,discard_unknown_columns"` // nolint
 
+	// SequenceNum is a unique identifier used for pagination
```
joostjager, line 42: I see now that there is no code required to just insert sequence numbers in the database, making this commit minimal. When suggesting to split, I somehow thought it would be more substantial. But if this is it, then it's it.

persistence/pg.go
```diff
@@ -81,7 +83,10 @@ func unmarshallDbInvoice(invoice *dbInvoice) *Invoice {
 				Value:           lnwire.MilliSatoshi(invoice.AmountMsat),
 			},
 		},
-		Settled: invoice.Settled,
+		SequenceNum:       invoice.SequenceNum,
+		Settled:           invoice.Settled,
+		SettleRequestedAt: invoice.SettleRequestedAt,
```
joostjager, line 88: Hmm an earlier omission?

persistence/pg.go
```diff
@@ -134,6 +139,27 @@ func (p *PostgresPersister) Get(ctx context.Context, hash lntypes.Hash) (*Invoic
 	return invoice, htlcs, nil
 }
 
+func (p *PostgresPersister) GetInvoices(ctx context.Context, maxInvoicesCount, sequenceOffset int) ([]*Invoice, error) {
+	var dbInvoices []*dbInvoice
```
joostjager, line 145: Newline above?

mux_test.go
```diff
@@ -382,6 +388,15 @@ func TestMux(t *testing.T) {
 
 	expectResponse(<-l[0].responseChan, 20, routerrpc.ResolveHoldForwardAction_SETTLE)
 
+	// Should have 2 invoices in the response (one settled and one to be settled)
```
joostjager: Assert `Settled` property then?

lnmuxrpc/lnmux.proto
```diff
@@ -73,3 +77,24 @@ message CancelInvoiceRequest {
 }
 
 message CancelInvoiceResponse {}
+
+message ListInvoicesRequest {
+    uint64 max_invoices_count = 1;
+    uint64 sequence_offset = 2;
```
joostjager: Nit: `sequence_start`? Offset still sounds a bit vague as to what it going to happen. But perhaps just me.

persistence/pg.go
Torakushi: Hmmm, I will check because sequence_num is `SERIAL`

persistence/pg.go
Torakushi: Yep, it seems (I think we didn't need it)

persistence/pg.go
Torakushi: Could be nice to add a linter for that

persistence/pg.go
joostjager: I think go-pg doesn't know about that and just tries to map to NULL. We are never setting this number, so it probably doesn't matter.

persistence/pg.go
joostjager: https://github.com/bottlepay/lnmux/pull/120

persistence/pg.go
Torakushi: I left it like that (If you really prefer, I will change it)

joostjager: Yes, can leave as is.

cmd/lnmuxd/server.go
```diff
@@ -294,3 +298,46 @@ func (s *server) CancelInvoice(ctx context.Context,
 
 	return &lnmuxrpc.CancelInvoiceResponse{}, nil
 }
+
+func (s *server) ListInvoices(ctx context.Context, req *lnmuxrpc.ListInvoicesRequest) (*lnmuxrpc.ListInvoicesResponse, error) {
+	if req.MaxInvoicesCount == 0 {
+		return nil, status.Errorf(codes.InvalidArgument, "should provide 'num_max_invoices' field")
```
joostjager: update error message field name

joostjager:
Tested and works

Torakushi:
I didn't mention it but I also tested it :)

lnmuxrpc/lnmux.proto
```diff
@@ -19,6 +19,10 @@ service Service {
     // Cancels an accepted invoice. In case settle has been requested
     // for an invoice, CancelInvoice returns a FailedPrecondition error.
     rpc CancelInvoice (CancelInvoiceRequest) returns (CancelInvoiceResponse);
+
+    // Lists settled or to be settled invoices. It returns invoices with a sequence number between
+    // 'index_offset' and 'index_offset + num_max_invoices'.
```
joostjager: Update comment.

joostjager (approved):
One nit left

lnmuxrpc/lnmux.proto
Torakushi: Sorry, Should have re-read more carefully

joostjager: It still says `index_offset` 😛

Torakushi: Correction: I WILL re-read my PR 😄

persistence/pg.go
```diff
@@ -131,6 +139,29 @@ func (p *PostgresPersister) Get(ctx context.Context, hash lntypes.Hash) (*Invoic
 	return invoice, htlcs, nil
 }
 
+func (p *PostgresPersister) GetInvoices(ctx context.Context,
+	maxInvoicesCount, sequenceOffset int) ([]*Invoice, error) {
+
+	var dbInvoices []*dbInvoice
+	err := p.conn.RunInTransaction(ctx, func(tx *pg.Tx) error {
```
stanislavkozlovski: does a single select need to run in a transaction?

Torakushi: Thanks for reviewing!
Indeed we gain nothing from a transaction here

Torakushi: I will merge for now, but if you have any other concerns about this PR, feel free to comment. I will be more than happy to address them!

PR #106, opened 2022-11-01
Use final settle signal

### LINKED ISSUE #105: Use final settle signal from lnd
With the merge of https://github.com/lightningnetwork/lnd/pull/6517, it is now possible to wait for the newly introduced final settle signal before marking an invoice as settled in lnmux.

joostjager:
Fixes #105


node_settled_handler.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	// Retrieve all htlcs that are not yet settled.
+	htlcs, err := p.persister.GetPendingHtlcs(ctx, p.lnd.PubKey())
+	if err != nil {
+		return err
+	}
+
+	// Look up each htlc to see if it has been settled in the mean time.
+	for key := range htlcs {
```
Torakushi, line 83: Good catch

node_settled_handler.go
```diff
@@ -0,0 +1,156 @@
...
+		if err != nil {
+			return err
+		}
+
+		finalEvent, ok := event.Event.(*routerrpc.HtlcEvent_FinalHtlcEvent)
+		if !ok {
+			continue
+		}
+		if !finalEvent.FinalHtlcEvent.Settled {
+			// TODO: Handle final fail?
```
joostjager, line 116: @Torakushi what do you think we should do in this case? The user of lnmux requested to settle, but lnd lost the payment. The exact scenario where we are doing this for. Current behavior is I think for the invoice to remain in settle requested indefinitely.

mux_test.go
```diff
@@ -327,6 +365,9 @@ func TestMux(t *testing.T) {
 	// Expect a settle signal for node 2.
 	expectResponse(<-l[1].responseChan, 2, routerrpc.ResolveHoldForwardAction_SETTLE)
 
+	// Notify final htlc resolution for htlc 2 on node 2.
+	l[1].notifyFinal(2)
+
 	// Wait for invoice-level settle signal.
 	settledChan := make(chan struct{})
 	go func() {
```
joostjager, line 389: I saw one CI test failure. Timeout on this goroutine.

https://github.com/bottlepay/lnmux/actions/runs/4364502046/jobs/7631898983

joostjager: Turned out to be a test problem with a `time.Sleep` that wasn't long enough. Replaced by a proper wait for signal.

lnd/client.go
```diff
@@ -151,6 +163,13 @@ func (l *lndClient) tryValidateConfig(ctx context.Context) error {
 		return ErrInterceptorNotRequired
 	}
 
+	// Check to see if lnd is running with ----store-final-htlc-resolutions.
+	// Otherwise we can't obtain final resolution information and invoices will
+	// never be marked as settled.
+	if !info.StoreFinalHtlcResolutions {
+		return ErrFinalHtlcResolutionsNotStored
```
Torakushi, line 172: Maybe we should put the indication in the README?

node_settled_handler.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+func (p *NodeSettledHandler) Run(ctx context.Context) {
+	p.logger.Infow("Starting node settled handler")
+
+	for {
+		err := p.subscribeEvents(ctx)
+		switch {
+		// TODO: Re-enable for no logging.
+		// case err == context.DeadlineExceeded:
```
Torakushi: Could be nice to enable this case with a `continue` (to lose the least amount of event possible)

node_settled_handler.go
```diff
@@ -0,0 +1,156 @@
...
+	Lnd             lnd.LndClient
+	SettledCallback func(lntypes.Hash)
+}
+
+type NodeSettledHandler struct {
+	persister *persistence.PostgresPersister
+	logger    *zap.SugaredLogger
+	lnd       lnd.LndClient
+
+	finalHtlcsChan chan types.CircuitKey
```
Torakushi: Not used I think? (same for the one bellow)

node_settled_handler.go
Torakushi: IMO (but maybe I am missing something). 
The user requested to settle the payment, so we've added the invoice + HTLCs in the DB (invoice hash is a PK).
If we receive a `FinalHtlcEvent` `with `settled = false` (MalformedHTLC, UpdateFailHTLC,..) then, there is no way for the associated invoice to be settled now because the HTLC set will never be completed. 

So, I think the invoice should be marked as fail? (Maybe add an other column in the invoice/htlc table like `STATE` than can be `SETTLED` or `FAILED`, or simply delete the invoice + associated HTLCs?)

node_settled_handler.go
jjager-nydig: Done. Also added a context check to the retry delay.

lnd/client.go
jjager-nydig: Added

node_settled_handler.go
jjager-nydig: Indeed, removed.

node_settled_handler.go
joostjager: Discussed offline. Conclusion: we should indeed move the invoice to a final state always (either settled or failed). And also bubble up the outcome via the `WaitForSettled` rpc (which may need renaming to `WaitForFinal`).

node_settled_handler.go
```diff
@@ -0,0 +1,166 @@
...
+	Lnd             lnd.LndClient
+	SettledCallback func(lntypes.Hash)
+}
+
+type NodeSettledHandler struct {
+	persister *persistence.PostgresPersister
+	logger    *zap.SugaredLogger
+	lnd       lnd.LndClient
+
+	newWatchChan chan types.CircuitKey
```
Torakushi: I think it's unused?

node_settled_handler.go
Torakushi: https://github.com/bottlepay/lnmux/pull/131

.github/workflows/test.yml
Torakushi, line 1: Question, do we want to keep this behaviour in our pipeline (could be nice do detect flaky tests)

Torakushi:
Otherwise LGTM!

.github/workflows/test.yml
Torakushi: (Maybe I've misinterpreted the `[dev]` flag in commit, did you plan to remove this commit?)

node_settled_handler.go
joostjager: Removed

.github/workflows/test.yml
joostjager: Yes, was planning to remove. Can re-add it if we have another flakey test.

Torakushi: IMO, it would be good to add it. Not sure that one always thinks to add `--count` locally to check if there is any flaky test.

what do you think?

EDIT: NO

Torakushi (approved):

PR #107, opened 2022-11-02
Add node to key

joostjager:
Fixes bug where two nodes connected to multiplexer share a channel and cause circuit key collisions.

Circuit format in logs: `circuit=02a508a76083ef51bd60c994f5756e16aaafa946a29886397f411f3da0b32aa6ea:4039605720580096:19`

### DIFF
--- a/set_id.go
+++ b/set_id.go
@@ -1,6 +1,7 @@
 package lnmux
 
 import (
+	"bytes"
 	"crypto/sha256"
 	"encoding/hex"
 	"sort"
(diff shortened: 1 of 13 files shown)


set_id.go
```diff
@@ -16,16 +17,20 @@ func (s SetID) String() string {
...
-func newSetID(keys []types.CircuitKey) SetID {
-	sortedKeys := make([]types.CircuitKey, len(keys))
+func newSetID(keys []types.HtlcKey) SetID {
+	sortedKeys := make([]types.HtlcKey, len(keys))
 	copy(sortedKeys, keys)
 
 	sort.Slice(sortedKeys, func(a, b int) bool {
-		if sortedKeys[a].ChanID == sortedKeys[b].ChanID {
-			return sortedKeys[a].HtlcID < sortedKeys[b].HtlcID
+		if sortedKeys[a].Node != sortedKeys[b].Node {
```
joostjager, line 25: Fixing the hash collision!

Torakushi: What happens if `len(keys)==1`?

we could still have the same set_id if we make a payment A-->B or B-->A, do we want this behaviour?

Torakushi: I mean, in case of self-payment:

NewSetID(nodeA,htlcID,channelID) == NewSetID(nodeB,htlcID,channelID)

joostjager: Oops, forgot to add the node key 😬

joostjager: Also changed it to use a writer, to avoid the indexing logic.

set_id_test.go
```diff
@@ -4,26 +4,59 @@ import (
...
+	hash = newSetID(keys)
+	require.NotEqual(t, expectedHash, hash[:])
+
+	keys = testKeys()
 	keys[0].HtlcID = 5
 	hash = newSetID(keys)
 	require.NotEqual(t, expectedHash, hash[:])
 }
+
+// TestSetIDSelfPayments tests that two payments in different directions between
```
Torakushi, line 46: Awesome!

Torakushi (approved):
LGTM!

PR #109, opened 2022-11-03
bump lnd dependency to v0.15.4-beta

joostjager:
Intermediate step for a further bump to lnd 0.16 which will contain the final htlc api.


Torakushi (approved):
LGTM

PR #110, opened 2022-11-03
idempotent htlc settle

joostjager:
Fixes a bug where the interceptor would keep breaking on the error that was returned when the htlc was already settled in the database.

### DIFF
--- a/persistence/pg.go
+++ b/persistence/pg.go
@@ -162,7 +162,7 @@ func (p *PostgresPersister) RequestSettle(ctx context.Context,
 }
 
 func (p *PostgresPersister) MarkHtlcSettled(ctx context.Context,
-	hash lntypes.Hash, key types.HtlcKey) (bool, error) {
+	key types.HtlcKey) (bool, error) {
 
 	var invoiceSettled bool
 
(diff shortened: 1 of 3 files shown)


persistence/pg.go
```diff
@@ -175,18 +178,35 @@ func (p *PostgresPersister) MarkHtlcSettled(ctx context.Context,
...
 		result, err := tx.ModelContext(ctx, &htlc).
 			WherePK().
-			Where("hash=?", hash).
 			Set("settled=?", true).
 			Set("settled_at=?", now).
 			Update() // nolint:contextcheck
 		if err != nil {
 			return err
 		}
-		if result.RowsAffected() == 0 {
```
Torakushi, line 187: Why removing this?

Torakushi (approved):
Straight forward change
LGTM

joostjager:
@Torakushi I moved extracting the hash up: https://github.com/bottlepay/lnmux/compare/84ad5f0e0dc69f0469c6fe4c821ea390e71c1701..1286a795889a73702624009d4a1557ae0a15010e

I think it is not strictly necessary, but feels a bit safer.

PR #111, opened 2022-11-03
add node to primary key

joostjager:
Oversight in #107 

Impacts `MarkHtlcSettled` by incorrectly marking a potential other htlc as settled too.


Torakushi (approved):
LGTM

PR #112, opened 2022-11-03
one mux instance per lnd node

joostjager:
Keep all per-node logic together so that any failures remain isolated to that node. This PR also prepares for a potential more distributed future for lnmux.

### DIFF
--- a/cmd/lnmuxd/run.go
+++ b/cmd/lnmuxd/run.go
@@ -173,20 +173,6 @@ func run(ctx context.Context, cfg *Config) error {
 		},
 	)
 
-	mux, err := lnmux.New(
-		&lnmux.MuxConfig{
-			KeyRing:         keyRing,
-			ActiveNetParams: activeNetParams,
-			Lnd:             lnds,
-			Logger:          log,
-			Registry:        registry,
-			SettledHandler:  settledHandler,
-			RoutingPolicy:   routingPolicy,
-		})
-	if err != nil {
-		return err
-	}
-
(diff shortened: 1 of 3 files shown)


mux_test.go
```diff
@@ -378,6 +385,9 @@ func TestMux(t *testing.T) {
 	cancelAcceptSubscription()
 
 	cancel()
+
+	require.NoError(t, <-errChan)
```
Torakushi, line 390: Just for the reader (As I was confused at first)

Add a comment saying one is for registry and the two others are for mux clients

Torakushi (approved):
Didn't expect it will be that straight forward! This is cool!

LGTM

mux_test.go
joostjager: Added

cmd/lnmuxd/run.go
```diff
@@ -249,18 +235,47 @@ func run(ctx context.Context, cfg *Config) error {
...
 	}
 
-	group, ctx := errgroup.WithContext(ctx)
-
 	// Run multiplexer.
-	group.Go(func() error {
-		err := mux.Run(ctx)
+	var muxes []*lnmux.Mux
+	for _, lnd := range lnds {
+		mux, err := lnmux.New(
```
joostjager, line 241: @Torakushi, I made a change to first instantiate all muxes and only then spin up the goroutines. Otherwise we may have lingering goroutines if there is an error.

Torakushi: Nice catch! Remind me @91 :)

PR #115, opened 2022-11-04
persistence: fix race condition in MarkHtlcSettled

### LINKED ISSUE #114: Concurrency issue in MarkHtlcSettled
It seems that this line:

https://github.com/bottlepay/lnmux/blob/5e53736a809912877aafb102c4b8c6d7968882db/persistence/pg.go#L212

is not safe when multiple calls to MarkHtlcSettled execute concurrently. In a two-htlc set, both htlcs may see that the other htlc is still not settled, and both will conclude that the invoice is not yet settled while it in fact is.

joostjager:
fixes #114


Torakushi (approved):
Straight forward
nice catch!

PR #116, opened 2022-11-07
persistence: replace table lock by invoice row lock

joostjager:
Improve performance by avoiding a table-wide lock.

### DIFF
--- a/persistence/pg.go
+++ b/persistence/pg.go
@@ -90,9 +90,13 @@ func (p *PostgresPersister) Get(ctx context.Context, hash lntypes.Hash) (*Invoic
 	var invoice *Invoice
 
 	err := p.conn.RunInTransaction(ctx, func(tx *pg.Tx) error {
+		// Retrieve the invoice and lock it FOR SHARE. This is to prevent an
+		// update occurring before retrieving the htlcs in the next query.
 		var dbInvoice dbInvoice
 		err := tx.ModelContext(ctx, &dbInvoice).
-			Where("hash=?", hash).Select()
+			Where("hash=?", hash).
+			For("SHARE").
+			Select()
 		switch {
 		case err == pg.ErrNoRows:
 			return types.ErrInvoiceNotFound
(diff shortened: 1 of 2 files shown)


persistence/pg.go
```diff
@@ -186,9 +178,11 @@ func (p *PostgresPersister) MarkHtlcSettled(ctx context.Context,
 			HtlcID: key.HtlcID,
 		}
 
-		// Check to see if htlc exists.
-		err = tx.ModelContext(ctx, &htlc).
+		// Check to see if htlc exists. Use FOR UPDATE to prevent a concurrent
+		// call for the same htlc to execute in parallel.
+		err := tx.ModelContext(ctx, &htlc).
 			WherePK().
+			For("UPDATE").
```
joostjager: Postgres locking is not easy. Maybe the `Get` function also isn't completely safe if an update comes in between the two select statements in there. Might need to use repeatable read isolation.

joostjager: Maybe repeatable read/serializable doesn't require `FOR UPDATE` then here?

Torakushi: Regarding the safety, I agree with you, maybe we can acquire the lock for the invoice before acquiring the lock for the htlc? 

So we should not have problems?

Torakushi: Or

We could SELECT FOR UPDATE all htlcs linked to the invoice (But I really don't like this idea)

persistence/pg.go
```diff
@@ -90,9 +90,13 @@ func (p *PostgresPersister) Get(ctx context.Context, hash lntypes.Hash) (*Invoic
 	var invoice *Invoice
 
 	err := p.conn.RunInTransaction(ctx, func(tx *pg.Tx) error {
+		// Retrieve the invoice and lock it FOR SHARE. This is to prevent an
+		// update occuring before retrieving the htlcs in the next query.
 		var dbInvoice dbInvoice
 		err := tx.ModelContext(ctx, &dbInvoice).
-			Where("hash=?", hash).Select()
+			Where("hash=?", hash).
+			For("SHARE").
```
joostjager, line 98: Added FOR SHARE here.

persistence/pg.go
```diff
@@ -164,57 +168,81 @@ func (p *PostgresPersister) RequestSettle(ctx context.Context,
...
+func (p *PostgresPersister) getHtlcHash(ctx context.Context, tx *pg.Tx,
+	key types.HtlcKey) (lntypes.Hash, error) {
+
+	htlc := dbHtlc{
+		Node:   key.Node,
+		ChanID: key.ChanID,
+		HtlcID: key.HtlcID,
+	}
+
+	// Check to see if htlc exists. Use FOR UPDATE to prevent a concurrent
```
Torakushi: We need to update this comment

joostjager: Removed.

Torakushi (approved):
LGTM! Good work!

PR #117, opened 2022-11-07
use global lnd client timeouts

joostjager:
Clients should always guard their own timeouts. Additionally a standard timeouts helps to pull stream timeouts into the happy flow and surfaces potential issues sooner.

### DIFF
--- a/cmd/lnmuxd/run.go
+++ b/cmd/lnmuxd/run.go
@@ -324,7 +324,6 @@ func initLndClients(ctx context.Context, cfg *LndConfig) ([]lnd.LndClient, *chai
 			Logger:       log,
 			PubKey:       pubkey,
 			Network:      network,
-			Timeout:      cfg.Timeout,
 		})
 		if err != nil {
 			return nil, nil, err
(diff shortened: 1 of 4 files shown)


cmd/lnmuxd/run.go
```diff
@@ -324,7 +324,6 @@ func initLndClients(ctx context.Context, cfg *LndConfig) ([]lnd.LndClient, *chai
 			Logger:       log,
 			PubKey:       pubkey,
 			Network:      network,
-			Timeout:      cfg.Timeout,
```
Torakushi, line 327: Would users find it useful to make timeouts configurable?

Torakushi (approved):
LGTM

cmd/lnmuxd/run.go
joostjager: I thought they would not. It would also be two timeouts to configure after this pr. What do you think?

Torakushi: I don't think that will be very useful.
If in the future people want it to be configurable, it will be an easy change to make anyway.

PR #118, opened 2022-11-08
remove source field from interceptor output

joostjager:
Next step in isolating per-node logic.


Torakushi (approved):
LGTM! We are moving!

PR #119, opened 2022-11-10
add lnmux_disconnected_nodes metric

### LINKED ISSUE #15: Add prometheus metrics

joostjager:
fixes #15

### DIFF
--- a/interceptor.go
+++ b/interceptor.go
@@ -12,11 +12,21 @@ import (
 	"github.com/bottlepay/lnmux/types"
 	"github.com/lightningnetwork/lnd/lnrpc/routerrpc"
 	"github.com/lightningnetwork/lnd/lntypes"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 	"go.uber.org/zap"
 )
 
 const (
 	resolutionQueueSize = 100
+)
+
+// disconnectedNodesGaugeMetric tracks the number of configured lnd nodes to
+// which we do not have a connection.
+var disconnectedNodesGaugeMetric = promauto.NewGauge(
+	prometheus.GaugeOpts{
+		Name: "lnmux_disconnected_nodes",
+	},
 )
 
 type preSendCallbackFunc func(context.Context, common.PubKey, queuedReply) error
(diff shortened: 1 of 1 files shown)


interceptor.go
```diff
@@ -50,6 +60,12 @@ func newInterceptor(lnd lnd.LndClient, logger *zap.SugaredLogger,
 func (i *interceptor) run(ctx context.Context) {
 	defer i.logger.Debugw("Exiting interceptor loop")
 
+	// Start in the disconnected state. We are not supposed to exit this
```
Torakushi, line 63: Nice catch indeed

interceptor.go
```diff
@@ -12,13 +12,23 @@ import (
...
 	"go.uber.org/zap"
 )
 
 const (
 	resolutionQueueSize = 100
 )
 
+// disconnectedNodesGaugeMetric tracks the number of configured lnd nodes to
+// which we do not have a connection.
+var disconnectedNodesGaugeMetric = promauto.NewGauge(
```
Torakushi, line 26: I don't know but: does it make sense to group this and the subscriber gauge in a separate file? (Not sure)

joostjager: I thought to keep it close to where it is used. Otherwise it is so 'global', even though it is still global now.

interceptor.go
```diff
@@ -129,6 +145,10 @@ func (i *interceptor) start(ctx context.Context) error {
 		}
 	}(ctx)
 
+	// We consider ourselves connected now.
+	disconnectedNodesGaugeMetric.Dec()
+	defer disconnectedNodesGaugeMetric.Inc()
```
joostjager, line 150: An alert does need to have some delay period, because we do periodic stream disconnects.

Torakushi (approved):
LGTM!

interceptor.go
Torakushi: Anyway, in prom/datadog, we use time sample (if lnmux_disconnected_nodes > 1 for 5 min then alert)

PR #128, opened 2023-03-08
lnd connection checks

joostjager:
Make sure we've always checked the lnd instance for a correct configuration and add a version check.

### DIFF
--- a/lnd/client.go
+++ b/lnd/client.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sync"
 
 	"github.com/bottlepay/lnmux/common"
 	"github.com/btcsuite/btcd/chaincfg"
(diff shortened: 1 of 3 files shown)


Torakushi (approved):
LGTM

lnd/client.go
```diff
@@ -72,38 +80,80 @@ func NewLndClient(ctx context.Context, cfg Config) (LndClient, error) {
...
+	}
+
+	// Request node info.
+	info, err := l.lnClient.GetInfo(ctx, &lnrpc.GetInfoRequest{})
+	if err != nil {
+		return err
+	}
+
+	// Clip off unofficial commit indication from version reported by lnd.
+	versionString := info.Version
```
Torakushi: Why not use verrpc? (so we don't depend on a particular string format)

joostjager: I can use verrpc, but then need to replicate the version comparison logic (major > major, major == major, major < major, minor ..., patch ...).

An in-between solution could be to use verrpc in combination with go-version for the comparison?

joostjager: Went for this one.

Torakushi: Ah yeah... Just checked.
Indeed it will work,  I would prefer that instead of relying to the lnd string format!

Torakushi: Really prefer that! good job

PR #130, opened 2023-03-09
build: golang 1.19

joostjager:
Concrete reason to bump was the use of `atomic.Bool` in the final settle pr #106. But it is a good idea to upgrade to current stable anyway.


Torakushi (approved):
Great, I didn't know they had added atomic.Bool

PR #131, opened 2023-03-22
Handle htlc loss when looking for final HTLC resolution

Torakushi:
When looking for a HTLC final resolution, it's possible that LND lost the HTLC and returns a `HtlcEvent_FinalHtlcEvent` with settled == false.

In this case, we will mark the HTLC and the invoice as 'FAILED' and we will notify the user using the `WaitForInvoiceFinalEvent` (Formerly WaitForInvoiceSettled)

To do that we:

- added more generic columns to track the invoice resolution
- added the invoice status in the `WaitForInvoiceFinalEventResponse`


persistence/pg.go
```diff
@@ -260,68 +262,97 @@ func (p *PostgresPersister) MarkHtlcSettled(ctx context.Context,
...
 
-		now := time.Now().UTC()
-
-		htlc := dbHtlc{
-			Node:   key.Node,
-			ChanID: key.ChanID,
-			HtlcID: key.HtlcID,
+		// It is possible that we received previously a HTLC that has
+		// failed for this invoice. In consequence, the invoice should
+		// have the FAILED status. However, a HTLC could have been settled
```
Torakushi: I am not sure if this is a real case, I could not find anything explicit in LND.

db.go
```diff
@@ -43,11 +43,6 @@ var (
 type Invoice struct {
 	types.InvoiceCreationData
 
-	Settled bool
-
-	// SettleDate is the exact time the invoice was settled.
-	SettleDate time.Time
```
joostjager: Can move this into a separate PR on master to get it out of the way

lnmuxrpc/lnmux.proto
```diff
@@ -94,7 +94,7 @@ message Invoice {
     bytes preimage = 2;
     uint64 amount_msat = 3;
     bool settled = 4;
-    uint64 settled_requested_at = 5;
+    uint64 settle_requested_at = 5;
```
joostjager, line 106: Does the deposit connector need to be updated too?

cmd/lnmuxd/server.go
```diff
@@ -336,9 +336,25 @@ func dbInvoiceToProto(invoice *persistence.Invoice) *lnmuxrpc.Invoice {
...
+	var status lnmuxrpc.InvoiceStatus
+	switch invoiceStatus {
+	case types.StatusSettleRequested:
+		status = lnmuxrpc.InvoiceStatus_SETTLE_REQUESTED
+	case types.StatusSettled:
+		status = lnmuxrpc.InvoiceStatus_SETTLED
+	case types.StatusFailed:
+		status = lnmuxrpc.InvoiceStatus_FAILED
+	default:
+		status = lnmuxrpc.InvoiceStatus_UNKNOWN
```
joostjager: Is it happening, UNKNOWN?

persistence/migrations/4_add_status.up.sql
```diff
@@ -0,0 +1,13 @@
...
+
+ALTER TABLE lnmux.htlcs ADD "status" VARCHAR;
+ALTER TABLE lnmux.htlcs ADD "resolution_at" TIMESTAMPTZ;
+
+
+ALTER TABLE lnmux.invoices DROP COLUMN "settled";
+ALTER TABLE lnmux.invoices DROP COLUMN "settled_at";
+
+ALTER TABLE lnmux.htlcs DROP COLUMN "settled";
+ALTER TABLE lnmux.htlcs DROP COLUMN "settled_at";
```
joostjager, line 34: Data migration from old to new column?

persistence/pg.go
```diff
@@ -21,8 +21,9 @@ type Invoice struct {
 
 	SequenceNum       uint64
 	SettleRequestedAt time.Time
-	SettledAt         time.Time
-	Settled           bool
+	ResolutionAt      time.Time
+	Status            types.Status
+	ReachFinalStatus  bool
```
joostjager: Maybe this can be a method `IsFinal()` on `types.Status`? It looks redundant to store in the database.

cmd/lnmuxd/server.go
```diff
@@ -155,15 +155,15 @@ func (s *server) SubscribeInvoiceAccepted(req *lnmuxrpc.SubscribeInvoiceAccepted
...
+	req *lnmuxrpc.WaitForInvoiceFinalEventRequest) (
+	*lnmuxrpc.WaitForInvoiceFinalEventResponse, error) {
 
 	hash, err := lntypes.MakeHash(req.Hash)
 	if err != nil {
 		return nil, err
 	}
 
-	err = s.settledHandler.WaitForInvoiceSettled(ctx, hash)
+	invoiceStatus, err := s.settledHandler.WaitForInvoiceSettled(ctx, hash)
```
joostjager: It's not renamed as promised by the commit message?

cmd/lnmuxd/server.go
```diff
@@ -174,7 +174,9 @@ func (s *server) WaitForInvoiceSettled(ctx context.Context,
 		return nil, status.Errorf(codes.Internal, err.Error())
 	}
 
-	return &lnmuxrpc.WaitForInvoiceSettledResponse{}, nil
+	return &lnmuxrpc.WaitForInvoiceFinalEventResponse{
+		InvoiceStatus: statusToProto(invoiceStatus),
```
joostjager: Is this strict enough when only `SETTLED` and `FAILED` are returned? Alternatively it could be a boolean. But maybe the consistency with other calls is a plus too.

cmd/lnmuxd/server.go
```diff
@@ -155,15 +155,15 @@ func (s *server) SubscribeInvoiceAccepted(req *lnmuxrpc.SubscribeInvoiceAccepted
 }
 
 func (s *server) WaitForInvoiceSettled(ctx context.Context,
```
joostjager: WaitForInvoiceFinalEvent?

settled_handler.go
```diff
@@ -64,22 +65,27 @@ func (p *SettledHandler) WaitForInvoiceSettled(ctx context.Context,
 	// Check database to see if invoice was already settled.
 	invoice, _, err := p.persister.Get(ctx, hash)
 	if err != nil {
-		return err
+		return types.StatusUnknown, err
```
joostjager: For enums that don't matter, I usually return `0, err`. To make it clear that it really does not matter what the primary return value is.

node_settled_handler.go
```diff
@@ -115,25 +110,60 @@ func (p *NodeSettledHandler) subscribeEvents(ctx context.Context) error {
...
+
+		if err := p.handleFinalHtlc(ctx, key, failedResolution); err != nil {
 			return err
 		}
 	}
 }
 
 func (p *NodeSettledHandler) handleFinalHtlc(ctx context.Context,
-	key types.CircuitKey) error {
+	key types.CircuitKey, failedResolution bool) error {
```
joostjager: Mostly `settled bool` is used, but this is `failed bool`. Maybe better to keep it consistent and always use `settled`.

node_settled_handler.go
```diff
@@ -115,25 +110,60 @@ func (p *NodeSettledHandler) subscribeEvents(ctx context.Context) error {
...
+		}
+
+		status = types.StatusFailed
+	} else {
+		invoiceHash, err = p.handleFinalSettledHtlc(ctx, key)
+		if err != nil {
+			return err
+		}
+
+		status = types.StatusSettled
```
joostjager: If this is just for logging, maybe you can make a logger at the top of this function using `logger.With`, and then log inside the if blocks `Infow("final final received")` and `Infow("final settle received")`?

node_settled_handler.go
```diff
@@ -143,24 +173,43 @@ func (p *NodeSettledHandler) handleFinalHtlc(ctx context.Context,
 
 	settledHash, err := p.persister.MarkHtlcSettled(ctx, htlcKey)
 	switch {
-	case err == persistence.ErrHtlcAlreadySettled:
-		return nil
+	case err == persistence.ErrHtlcAlreadyInStatus:
```
joostjager: I see you want to reuse the err message, but this name isn't optimally clear I think. Perhaps `ErrHtlcAlreadyFinal`?

node_settled_handler.go
```diff
@@ -143,24 +173,43 @@ func (p *NodeSettledHandler) handleFinalHtlc(ctx context.Context,
...
-	if settledHash != nil {
-		p.settledCallback(*settledHash)
+	htlcKey := types.HtlcKey{
+		ChanID: key.ChanID,
+		HtlcID: key.HtlcID,
+		Node:   p.lnd.PubKey(),
 	}
 
-	return nil
+	failedHash, err := p.persister.MarkHtlcFailed(ctx, htlcKey)
```
joostjager: Instead of this duplication, maybe you can do a `persister.MarkHtlcFinal(ctx, htlckey, settled)`?

persistence/pg.go
```diff
@@ -331,15 +362,66 @@ func (p *PostgresPersister) MarkHtlcSettled(ctx context.Context,
...
 	})
 	if err != nil {
 		return nil, err
 	}
 
-	return settledHash, nil
+	return invoiceHash, nil
+}
+
+func (p *PostgresPersister) selectInvoiceForUpdate(ctx context.Context,
```
joostjager, line 335: Refactor can go to a separate commit?

persistence/pg.go
```diff
@@ -260,68 +262,97 @@ func (p *PostgresPersister) MarkHtlcSettled(ctx context.Context,
...
-		htlc := dbHtlc{
-			Node:   key.Node,
-			ChanID: key.ChanID,
-			HtlcID: key.HtlcID,
+		// It is possible that we received previously a HTLC that has
+		// failed for this invoice. In consequence, the invoice should
+		// have the FAILED status. However, a HTLC could have been settled
+		// afterward.
+		if invoice.Status == types.StatusFailed {
+			return nil
```
joostjager: By returning `nil` another notification is prevented?

persistence/pg.go
```diff
@@ -260,68 +262,97 @@ func (p *PostgresPersister) MarkHtlcSettled(ctx context.Context,
...
+			Where("hash=?", invoice.Hash).
 			Set("status=?", types.StatusSettled).
 			Set("resolution_at=?", now).
 			Update() // nolint:contextcheck
 		if err != nil {
 			return err
 		}
 		if result.RowsAffected() == 0 {
-			return ErrHtlcAlreadySettled
+			return types.ErrInvoiceNotFound
```
joostjager: I thought at this point the invoice should always be there?

settled_handler.go
```diff
@@ -64,22 +65,27 @@ func (p *SettledHandler) WaitForInvoiceSettled(ctx context.Context,
...
+		return invoice.Status, nil
 	}
 
 	// Not settled yet. Wait for the event.
 	select {
 	case <-waitChan:
 		logger.Debugw("Wait for invoice settled completed via wait channel")
 
-		return nil
+		invoice, _, err := p.persister.Get(ctx, hash)
```
joostjager: Is this re-fetch necessary or can the final outcome be passed over the `waitChan`?

persistence/pg.go
joostjager: I think this is real.

lnmuxrpc/lnmux.proto
Torakushi: No, this `message` is only use for the ListInvoices rpc call

persistence/migrations/4_add_status.up.sql
Torakushi: What do you mean exactly?

I removed settled and settled_at in both invoices and htlcs tables and added
status and resolution_at in both tables?

the migration down is done in 4_add_status.down.sql

persistence/pg.go
Torakushi: `ReachFinalStatus` is never stored in the DB (we use the `dbInvoice` structure to store in DB)
`ReachFinalStatus` is only a field to know when the invoice reaches its final status.

But I do agree, having IsFinal() is better than a new field!

cmd/lnmuxd/server.go
Torakushi: oh.... nice catch

cmd/lnmuxd/server.go
Torakushi: I didn't want to rename it in the `settledHandler` handler because (firstly), I thought that it is normal for a settledHandler to wait for settled 😄 

And, I also thought that, the failed status is a specific case occurring when we wait for the settled signal.

But I would be quite happy to rename it here too!

cmd/lnmuxd/server.go
Torakushi: Hmm you are right. In this case, we are only waiting for SETTLED or FAILED,

I will restrict `statusToProto` cases.

Personaly, I would prefer keeping this enum (for consistency like you said). But I can totally make it a boolean!

cmd/lnmuxd/server.go
Torakushi: linked to this comment: https://github.com/bottlepay/lnmux/pull/131/files#r1144875714

I will restrict the cases here.

Unknown is only return by the settledHandler in case of error

persistence/pg.go
Torakushi: By returning nil, we avoid to mark an invoice as settled if it has been already marked as failed.

Do you prefer to return an error and add a log in the caller?

persistence/pg.go
Torakushi: The github diff is misleading

`return types.ErrInvoiceNotFound` is already present in settle_final branch
for the invoice update:

https://github.com/bottlepay/lnmux/pull/106/files#diff-fa1f5d8279c623bad5b14b09e9cabf2c5121065152f2aaab652bb71a85fab6b7R329

But I can change it if you prefer?

settled_handler.go
Torakushi: First, I didn;t want to because It will need a bit of refactoring
(change 	`settledCallback func(lntypes.Hash)` in node_settle_handler.go, mux.go,...

But I can if you think that the re-fetch is really annoying!

lnmuxrpc/lnmux.proto
joostjager: Np then :+1:

cmd/lnmuxd/server.go
joostjager: On second thought, I think it's fine to keep the enum.

cmd/lnmuxd/server.go
joostjager: Ah, so that will never be mapped to an actual proto enum. I think you can remove it then?

persistence/migrations/4_add_status.up.sql
joostjager: I mean migrating the data that is already present in those tables.

cmd/lnmuxd/server.go
joostjager: I mean `WaitForInvoiceSettled` :)

persistence/pg.go
joostjager: Anything refactor can be pulled out of the PR and merged onto master, if it makes sense.

persistence/pg.go
joostjager: Ah no, it's ok. I was just checking to see what the effect of `nil` is here. But tracing through the code, I see that ultimately any rpc waiters for settled aren't notified. And this seems good, because the final event already happened earlier.

persistence/pg.go
joostjager: Ah ok. And what needs to happen to end up here? The invoice is in the database, that has been checked in `selectInvoiceForUpdate`. If it is never supposed to happen, I'd return a different error than `ErrInvoiceNotFound`.

settled_handler.go
joostjager: I'd just do the refactor. I think it also helps with understanding the flow.

settled_handler.go
Torakushi: Never thought of that, nice one, will go for `"", err`

persistence/pg.go
Torakushi: We will be here if the invoice has state has changed in the meantime (not SETTLE_REQUESTED anymore). So we can't find the invoice with the given state

Should never happen (because we did a selectForUpdate).

cmd/lnmuxd/server.go
Torakushi: On a second thought as well 😄 : We need to keep `SETTLE_REQUESTED` in the statusToProto because of the `ListInvoices` RPC call

node_settled_handler.go
Torakushi: now that the wait channel will return the invoice status, it's not only for logging

node_settled_handler.go
```diff
@@ -29,7 +29,7 @@ type NodeSettledHandler struct {
 
 	newWatchChan chan types.CircuitKey
 
-	settledCallback func(lntypes.Hash)
+	settledCallback func(lntypes.Hash, types.Status)
```
Torakushi: should we rename `settledCallback` as well? IMO, it's not needed but... it's me 😄

persistence/pg.go
```diff
@@ -355,6 +351,40 @@ func (p *PostgresPersister) settleHTLC(ctx context.Context,
 	return nil
 }
 
+func (p *PostgresPersister) changeInvoiceStatus(ctx context.Context,
```
Torakushi: should it be in a separate commit? (ideally it should have been in the refactoring PR..)

lnmuxrpc/lnmux.proto
```diff
@@ -89,12 +89,19 @@ message ListInvoicesResponse {
...
+
 message Invoice {
     bytes hash = 1;
     bytes preimage = 2;
     uint64 amount_msat = 3;
-    bool settled = 4;
+    InvoiceStatus status = 4;
     uint64 settle_requested_at = 5;
-    uint64 settled_at = 6;
+    uint64 resolution_at = 6;
```
joostjager: I have some slight doubts about this name, but also not sure about something better. Maybe `finalized_at`, but that's also not ideal.

`final_at`, `completed_at` maybe?

persistence/migrations/4_add_status.up.sql
joostjager: Looks good

persistence/pg.go
joostjager: I think it's better in a separate commit indeed.

persistence/migrations/4_add_status.up.sql
```diff
@@ -0,0 +1,23 @@
+-- Migrate invoices table
+ALTER TABLE lnmux.invoices ADD "status" VARCHAR;
```
joostjager: I think you can make an enum in postgres.

cmd/lnmuxd/server.go
```diff
@@ -154,16 +154,16 @@ func (s *server) SubscribeInvoiceAccepted(req *lnmuxrpc.SubscribeInvoiceAccepted
...
+	req *lnmuxrpc.WaitForInvoiceFinalEventRequest) (
+	*lnmuxrpc.WaitForInvoiceFinalEventResponse, error) {
 
 	hash, err := lntypes.MakeHash(req.Hash)
 	if err != nil {
 		return nil, err
 	}
 
-	err = s.settledHandler.WaitForInvoiceSettled(ctx, hash)
+	invoiceStatus, err := s.settledHandler.WaitForInvoiceFinalEvent(ctx, hash)
```
joostjager: `WaitForInvoiceFinalStatus`, to remain consistent with the variable name/status enum?

mux.go
```diff
@@ -36,7 +36,7 @@ type Mux struct {
 type MuxConfig struct {
 	KeyRing         keychain.SecretKeyRing
 	ActiveNetParams *chaincfg.Params
-	SettledCallback func(lntypes.Hash)
+	SettledCallback func(lntypes.Hash, types.Status)
```
joostjager: Maybe make it `types.InvoiceStatus`. Otherwise confusion will happen with the next status type added.

joostjager: `HtlcStatus` can be an alias, but it is still nice to have a separation between the types to prevent potential bugs.

node_settled_handler.go
```diff
@@ -115,51 +110,61 @@ func (p *NodeSettledHandler) subscribeEvents(ctx context.Context) error {
...
+		"chanID", key.ChanID,
+		"htlcID", key.HtlcID,
+		"hash", invoiceHash,
+		"status", status,
+	)
 
-	if settledHash != nil {
-		p.settledCallback(*settledHash, types.StatusSettled)
+	if invoiceHash != nil {
+		p.settledCallback(*invoiceHash, status)
```
joostjager: Yes, so here I wonder if it isn't better to use a boolean, to make it more strict?

persistence/pg.go
```diff
@@ -249,53 +251,77 @@ func (p *PostgresPersister) getHtlcHash(ctx context.Context, tx *pg.Tx,
...
 		if err != nil {
 			return err
 		}
-		if !allHtlcsSettled {
+
+		// It is possible that we received previously a HTLC that has
+		// failed for this invoice. In consequence, the invoice should
+		// have the FAILED status. However, a HTLC could have reached
+		// its final status afterward.
+		if invoice.Status == types.StatusFailed {
```
joostjager: Still think it is worth adding to the comment here why we don't return an invoice hash for a final invoice.

joostjager: Maybe also catch `StatusSettled` here explicitly and return an error? Something must be in an inconsistent state if this happens right?

persistence/pg.go
joostjager: So then it's better to return a different error? To not hide a sitaution that is never supposed to happen?

node_settled_handler.go
joostjager: For consistency yes, `finalCallback` or `completedCallback`?

persistence/pg.go
Torakushi: In which case this will happen?

I mean, if invoice has StatusSettled state, it means that all HTLC were settled and MarkHtlcFinal will never be called again for this invoice

joostjager: Yes, as a sanity check I mean

lnmuxrpc/lnmux.proto
Torakushi: Will work my naming skill

persistence/migrations/4_add_status.up.sql
Torakushi: Fair, I was thinking about this.

In that case, we may want to create 2 different enums, one for htlc status and the other one for invoice status
(to be consistent with https://github.com/bottlepay/lnmux/pull/131/commits/49a926f62e7eea351e309e97dace35ab011199bf#r1146231182)

lnmuxrpc/lnmux.proto
joostjager: Just ask your AI assistant to come up with options 😆

persistence/migrations/4_add_status.up.sql
joostjager: Yes, I'd opt for separate types.

mux.go
Torakushi: Yeah I do agree on the separation!

To be consistent with what we did in the DB, I will create 2 distinct types (not aliasing otherwise it can be confusing to have a `HTLCStatus` with  type `InvoiceStatus`)

It will ease things if one day we want to change/add a status for either invoice or htlc

node_settled_handler.go
Torakushi: After a second thought, maybe it makes more sense here indeed.
More strict, and it makes sense that `settleHandler.InvoiceSettled` sends a bool in the channel instead of a status.

Then the server, depending on the bool, will return the right status (I really prefer to send the status (and not the bool) through the RPC call, it will be confusing the the user otherwise)

node_settled_handler.go
```diff
@@ -115,51 +110,61 @@ func (p *NodeSettledHandler) subscribeEvents(ctx context.Context) error {
...
 			return err
 		}
 	}
 }
 
 func (p *NodeSettledHandler) handleFinalHtlc(ctx context.Context,
-	key types.CircuitKey) error {
+	key types.CircuitKey, settled bool) error {
+
+	var status types.HtlcStatus
```
Torakushi: It's only for logging indeed.

If it's annoying, I can either put the logs directly in persistence (maybe not as consistent), or remove some info from the logs (the status)

persistence/pg.go
```diff
@@ -249,53 +263,87 @@ func (p *PostgresPersister) getHtlcHash(ctx context.Context, tx *pg.Tx,
...
+			return ErrHtlcReceivedButInvoiceAlreadySettled
 		}
 
-		// The HTLC is settled, now we check if all other HTLCs
-		// are settled as well.
-		allHtlcsSettled, err := p.allHtlcsAreSettled(ctx,
-			tx, invoice.Hash)
+		now := time.Now().UTC()
+
+		err = p.changeHTLCStatus(ctx, tx, key, htlcFinalStatus, now)
```
joostjager: nit: change sounds like you can change it to anything. Maybe `finalizeHTLC(settled)/finalizeInvoice(settled)`?

node_settled_handler.go
```diff
@@ -115,51 +110,61 @@ func (p *NodeSettledHandler) subscribeEvents(ctx context.Context) error {
...
+		"chanID", key.ChanID,
+		"htlcID", key.HtlcID,
+		"hash", invoiceHash,
+		"status", status,
+	)
 
-	if settledHash != nil {
-		p.finalCallback(*settledHash, true)
+	if invoiceHash != nil {
+		p.finalCallback(*invoiceHash, settled)
```
joostjager, line 157: Do you want to add test coverage for the invoice failed case?

joostjager (commented):
Looks close, lots of improvements made!

persistence/pg.go
joostjager: I think this has been addressed now right? Can't recall the full context of this comment with the new push.

node_settled_handler.go
```diff
@@ -115,51 +110,61 @@ func (p *NodeSettledHandler) subscribeEvents(ctx context.Context) error {
...
-	// If the htlc is not found, the final resolution was for an htlc that isn't
-	// managed by lnmux.
+	// If the htlc is not found, the final resolution was for an htlc that
+	// isn't managed by lnmux.
 	case err == types.ErrHtlcNotFound:
 		return nil
 
 	case err != nil:
-		return fmt.Errorf("unable to mark htlc %v settled: %w", key, err)
+		return fmt.Errorf("unable to mark htlc %v as %s : %w",
```
joostjager: Remove space before colon

node_settled_handler.go
joostjager: I think you can just log the settled boolean which you already have instead of mapping to status just for logging/err?

node_settled_handler.go
Torakushi: very very weird

I did the tests:

https://github.com/bottlepay/lnmux/commit/2abdd53414e07d4ca53f850c91a24b2daab6b41d#diff-b39fc442f543ad6df386d0772fdbafafee01095292545626a39251403bec6b5cR458

But for some reason, when doing a rebase, I removed them...

joostjager: Oh oh.... maybe something else changed. You can diff to some revision from `git reflog`

Torakushi: I triple checked, only this test was "lost"!

I just added the test again

lnmuxrpc/lnmux.proto
```diff
@@ -89,12 +89,19 @@ message ListInvoicesResponse {
     uint64 last_sequence_number = 3;
 }
 
+enum InvoiceStatus {
+    UNKNOWN = 0;
+    FAILED = 1;
+    SETTLE_REQUESTED=2;
```
jjager-nydig: ultra-nit: spaces around equal sign

persistence/migrations/4_add_status.up.sql
```diff
@@ -0,0 +1,34 @@
+CREATE TYPE invoice_status as ENUM (
+    'SETTLE_REQUESTED',
+    'SETTLED',
+    'FAILED'
+);
+
+-- Migrate invoices table
+ALTER TABLE lnmux.invoices ADD "status" invoice_status;
```
jjager-nydig: NOT NULL here and for htlcs?

persistence/migrations/4_add_status.up.sql
```diff
@@ -0,0 +1,34 @@
...
+ALTER TABLE lnmux.invoices ADD "finalized_at" TIMESTAMPTZ;
+
+UPDATE lnmux.invoices SET
+        status = CASE WHEN settled = FALSE THEN 'SETTLE_REQUESTED'::invoice_status ELSE 'SETTLED'::invoice_status END,
+        finalized_at = CASE WHEN settled = TRUE THEN settled_at ELSE NULL END;
+
+ALTER TABLE lnmux.invoices DROP COLUMN "settled";
+ALTER TABLE lnmux.invoices DROP COLUMN "settled_at";
```
jjager-nydig, line 16: I think that postgres also has a RENAME COLUMN statement. But this works too.

persistence/pg.go
```diff
@@ -378,7 +412,7 @@ func (p *PostgresPersister) finalizeInvoice(ctx context.Context,
 		return err
 	}
 	if result.RowsAffected() == 0 {
-		return types.ErrInvoiceNotFound
+		return ErrInvoiceAlreadyFinal
```
jjager-nydig: Hmm, looking at this function in isolation, I am not sure about this change.

Maybe the update needs a where clause checking for `SettleRequested` too.

I suppose you can't distinguish between a non-existing invoice and an invoice that is already final in one query.

joostjager (commented):
Unintentionally reviewed as @jjager-nydig

persistence/migrations/4_add_status.up.sql
Torakushi: Indeed, I didn't know it exists. It could be convenient for the migrate_up, but for the migrate_down, we still need to use the DROP (or a more complex logic) because for invoice with 'FAILED' status, `finalized_at` will be not NULL but we don't want `settled_at` to be not null (I think) 

I think I will keep the drop for now (if really no objections)

persistence/pg.go
Torakushi: when calling `finalizeInvoice` we always do a SELECT FOR UPDATE, in the same transaction.

So, the only case where `result.RowsAffected() == 0` are:

- invoice already in invoiceFinalStatus state
- invoice was removed from the DB manually in the meantime

However, I think that it's important to add a Where clause (`WHERE status = SETTLE_REQUESTED`). Because, IMO (even if it should never happen), `finalizeinvoice` should not change the status of an invoice that has already reached it's final status.
(currently if invoice is marked as FAILED, and we call `finalizeinvoice(SETTLED)`, it will change the invoice final status)

I've added this where statement.
Did I miss the point? (if the goal is to really know if the invoice wasn't deleted in the meantime, we can re-do a SELECT statement. But we still have no guarantee, we only narrow the gap)

joostjager: Yes, I know that those outcomes are the only ones. But I mean that just looking at this function in isolation, the context with select-for-update isn't visible.

Maybe it is enough to just rename the error to something that is always true for this function. Which is inevitably something more generic such as `ErrFinalizationFailed`. But then at least you can read this function without assumptions about the external environment.

Torakushi: ahhh, didn't understand that, will do!

joostjager (approved):

joostjager:
One more question, did you manually test the functionality with `lnd` nodes? This can be done by bringing down nodes at the right time, so that the peer is going to claw back the htlc and the invoice is going to end up in the failed state in the lnmux db.

Torakushi:
Yes! tested an checked

PR #132, opened 2023-03-22
code refactoring

Torakushi:
This PR is a quick refactor:

- remove unused code
- rename variable (typo)
- extract some code to prepare future development


cmd/lnmuxd/server.go
joostjager, line 1: LNMUX in the PR title isn't very useful for PRs in the lnmux repo... 😃

persistence/pg.go
```diff
@@ -301,6 +287,28 @@ func (p *PostgresPersister) MarkHtlcSettled(ctx context.Context,
...
+	if err != nil {
+		return nil, err
+	}
+
+	var invoice dbInvoice
+	err = tx.ModelContext(ctx, &invoice).
+		Where("hash=?", hash).
+		For("UPDATE").
+		Select()
+	switch {
```
joostjager, line 287: This can be an if statement?

Or you are missing the `err != nil` case?

persistence/pg.go
joostjager, line 1: Didn't you also want to extract that htlc update db code?

persistence/pg.go
Torakushi: Could be a if indeed,

But I will use a switch statement with another err!=nil to be consistent with the other part

persistence/pg.go
Torakushi: I was wondering

If I extract the function right now, I will need to modify its signature anyway afterward (to use 'status')

But maybe, it's more consistent

persistence/pg.go
```diff
@@ -285,8 +247,8 @@ func (p *PostgresPersister) MarkHtlcSettled(ctx context.Context,
 		}
 
 		// If all htlcs are settled, mark the invoice as settled.
-		_, err = tx.ModelContext(ctx, (*dbInvoice)(nil)).
-			Where("hash=?", hash).
+		result, err := tx.ModelContext(ctx, (*dbInvoice)(nil)).
```
Torakushi, line 250: There was a bug here

persistence/pg.go
Torakushi: done

persistence/pg.go
joostjager: Heh, nice one

persistence/pg.go
joostjager: I think it was a refactor bug previously without the case err != nil?

persistence/pg.go
joostjager: I think it still reduces the delta in the next pr, so probably worth it.

persistence/pg.go
Torakushi: It was not a bug because it was:

```
var invoice Invoice

err := SelectForUpdate(invoice)
switch{
       case err == pg.NoRows....
}

return invoice, err

```

So if err !=nil, the error was returned and checked by the caller

But it was not very clean..

joostjager (approved):

persistence/pg.go
joostjager: Ah yes, correct but dirty 😄

PR #140, opened 2024-01-09
fix 'gosec' linter error in tests

Torakushi:
gosec linter detects a "hardcoded password in a URL". We have indeed added a fallback url for PG instances in the test to be more practical.

We add `//nolint: gosec` for the given line to fix github's actions. (otherwise we can force the user to set the `LNMUX_TEST_DB_DSN` environment variable)


aakselrod (approved):
LGTM, it's for testing only.

